PR #4, opened 2021-05-21
[GEP-15] Add SSH command

xrstf:

Fixes #

**Release note**:

```feature user
Add `ssh` command. You can optionally pass the node name as argument
- Node name is provided : A remote shell to the node is opened automatically
- Node name is not provided: The bastion is kept alive and the `ssh` command to open a remote shell to a node, together with the list of the shoot nodes is printed to `stdout` for your convenience. You can then run the `ssh` command on your own
```

### DIFF
--- a/internal/fake/factory.go
+++ b/internal/fake/factory.go
@@ -41,12 +41,6 @@ type Factory struct {
 
 var _ util.Factory = &Factory{}
 
-func NewFakeManagerFactory(manager target.Manager) util.Factory {
-	return &Factory{
-		ManagerImpl: manager,
-	}
-}
-
 func NewFakeFactory(cfg *config.Config, clock util.Clock, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) *Factory {
 	if cfg == nil {
 		cfg = &config.Config{}
--- a/internal/fake/factory.go
+++ b/internal/fake/factory.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package fake
 
 import (
+	"context"
+
 	"github.com/gardener/gardenctl-v2/internal/util"
 	"github.com/gardener/gardenctl-v2/pkg/config"
 	"github.com/gardener/gardenctl-v2/pkg/target"
@@ -80,10 +82,14 @@ func (f *Factory) Manager() (target.Manager, error) {
 	return target.NewManager(f.Config, f.TargetProviderImpl, f.ClientProviderImpl, f.KubeconfigCacheImpl)
 }
 
-func (f *Factory) HomeDir() string {
+func (f *Factory) GardenHomeDir() string {
 	return f.GardenHomeDirectory
 }
 
 func (f *Factory) Clock() util.Clock {
 	return f.ClockImpl
+}
+
+func (f *Factory) PublicIP(ctx context.Context) (string, error) {
+	return "192.0.2.42", nil
 }
--- a/internal/fake/factory.go
+++ b/internal/fake/factory.go
@@ -16,6 +16,9 @@ import (
 
 // Factory implements util.Factory interface
 type Factory struct {
+	// ContextImpl is the root context any command should use.
+	ContextImpl context.Context
+
 	// Either set a specific Manager instance, or overwrite the
 	// individual providers/caches down below.
 	ManagerImpl target.Manager
@@ -44,13 +47,13 @@ func NewFakeManagerFactory(manager target.Manager) util.Factory {
 	}
 }
 
-func NewFakeFactory(cfg *config.Config, clock util.Clock, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) util.Factory {
+func NewFakeFactory(cfg *config.Config, clock util.Clock, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) *Factory {
 	if cfg == nil {
 		cfg = &config.Config{}
 	}
 
 	if clientProvider == nil {
-		clientProvider = NewFakeClientProvider(nil)
+		clientProvider = NewFakeClientProvider()
 	}
 
 	if kubeconfigCache == nil {
@@ -80,6 +83,14 @@ func (f *Factory) Manager() (target.Manager, error) {
 	}
 
 	return target.NewManager(f.Config, f.TargetProviderImpl, f.ClientProviderImpl, f.KubeconfigCacheImpl)
+}
+
+func (f *Factory) Context() context.Context {
+	if f.ContextImpl != nil {
+		return f.ContextImpl
+	}
+
+	return context.Background()
 }
 
 func (f *Factory) GardenHomeDir() string {
--- a/internal/fake/factory.go
+++ b/internal/fake/factory.go
@@ -101,6 +101,6 @@ func (f *Factory) Clock() util.Clock {
 	return f.ClockImpl
 }
 
-func (f *Factory) PublicIP(ctx context.Context) (string, error) {
-	return "192.0.2.42", nil
+func (f *Factory) PublicIPs(ctx context.Context) ([]string, error) {
+	return []string{"192.0.2.42", "2001:db8::8a2e:370:7334"}, nil
 }
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -7,8 +7,13 @@ SPDX-License-Identifier: Apache-2.0
 package util
 
 import (
+	"context"
 	"fmt"
+	"io/ioutil"
+	"net"
+	"net/http"
 	"path/filepath"
+	"strings"
 
 	"github.com/gardener/gardenctl-v2/pkg/config"
 	"github.com/gardener/gardenctl-v2/pkg/target"
@@ -18,10 +23,13 @@ import (
 type Factory interface {
 	// Clock returns a clock that provides access to the current time.
 	Clock() Clock
-	// HomeDir returns the home directory for the executing user.
-	HomeDir() string
+	// GardenHomeDir returns the home directory for the executing user.
+	GardenHomeDir() string
 	// Manager returns the target manager used to read and change the currently targeted system.
 	Manager() (target.Manager, error)
+	// PublicIP returns the current host's public IP address. It's
+	// recommended to provide a context with a timeout/deadline.
+	PublicIP(context.Context) (string, error)
 }
 
 // FactoryImpl implements util.Factory interface
@@ -56,10 +64,38 @@ func (f *FactoryImpl) Manager() (target.Manager, error) {
 	return target.NewManager(cfg, targetProvider, clientProvider, kubeconfigCache)
 }
 
-func (f *FactoryImpl) HomeDir() string {
+func (f *FactoryImpl) GardenHomeDir() string {
 	return f.GardenHomeDirectory
 }
 
 func (f *FactoryImpl) Clock() Clock {
 	return &RealClock{}
+}
+
+func (f *FactoryImpl) PublicIP(ctx context.Context) (string, error) {
+	req, err := http.NewRequest("GET", "https://api64.ipify.org/", nil)
+	if err != nil {
+		return "", err
+	}
+
+	req = req.WithContext(ctx)
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	ip, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	ipAddress := strings.TrimSpace(string(ip))
+
+	if net.ParseIP(ipAddress) == nil {
+		return "", fmt.Errorf("API returned an invalid IP (%q)", ipAddress)
+	}
+
+	return ipAddress, nil
 }
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -29,9 +29,11 @@ type Factory interface {
 	GardenHomeDir() string
 	// Manager returns the target manager used to read and change the currently targeted system.
 	Manager() (target.Manager, error)
-	// PublicIP returns the current host's public IP address. It's
-	// recommended to provide a context with a timeout/deadline.
-	PublicIP(context.Context) (string, error)
+	// PublicIP returns the current host's public IP addresses. It's
+	// recommended to provide a context with a timeout/deadline. The
+	// returned slice can contain IPv6, IPv4 or both, in no particular
+	// order.
+	PublicIPs(context.Context) ([]string, error)
 }
 
 // FactoryImpl implements util.Factory interface
@@ -78,30 +80,53 @@ func (f *FactoryImpl) Clock() Clock {
 	return &RealClock{}
 }
 
-func (f *FactoryImpl) PublicIP(ctx context.Context) (string, error) {
-	req, err := http.NewRequest("GET", "https://api64.ipify.org/", nil)
+func (f *FactoryImpl) PublicIPs(ctx context.Context) ([]string, error) {
+	ipv64, err := callIPify(ctx, "api64.ipify.org")
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+
+	addresses := []string{ipv64.String()}
+
+	// if the above resolved to IPv6, we also try the IPv4-only;
+	// this assumes everyone who has IPv6 also has IPv4
+	if ipv64.To4() == nil {
+		ipv4, err := callIPify(ctx, "api.ipify.org")
+		if err != nil {
+			return nil, err
+		}
+
+		addresses = append(addresses, ipv4.String())
+	}
+
+	return addresses, nil
+}
+
+func callIPify(ctx context.Context, domain string) (*net.IP, error) {
+	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s/", domain), nil)
+	if err != nil {
+		return nil, err
 	}
 
 	req = req.WithContext(ctx)
 
 	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 	defer resp.Body.Close()
 
 	ip, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	ipAddress := strings.TrimSpace(string(ip))
+	netIP := net.ParseIP(ipAddress)
 
-	if net.ParseIP(ipAddress) == nil {
-		return "", fmt.Errorf("API returned an invalid IP (%q)", ipAddress)
+	if netIP == nil {
+		return nil, fmt.Errorf("API returned an invalid IP (%q)", ipAddress)
 	}
 
-	return ipAddress, nil
+	return &netIP, nil
 }
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -88,15 +88,12 @@ func (f *FactoryImpl) PublicIPs(ctx context.Context) ([]string, error) {
 
 	addresses := []string{ipv64.String()}
 
-	// if the above resolved to IPv6, we also try the IPv4-only;
-	// this assumes everyone who has IPv6 also has IPv4
+	// if the above resolved to IPv6, we also _try_ the IPv4-only;
+	// this is optional and failures are silently swallowed
 	if ipv64.To4() == nil {
-		ipv4, err := callIPify(ctx, "api.ipify.org")
-		if err != nil {
-			return nil, err
+		if ipv4, err := callIPify(ctx, "api.ipify.org"); err == nil {
+			addresses = append(addresses, ipv4.String())
 		}
-
-		addresses = append(addresses, ipv4.String())
 	}
 
 	return addresses, nil
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -21,6 +21,8 @@ import (
 
 // Factory provides abstractions that allow the command to be extended across multiple types of resources and different API sets.
 type Factory interface {
+	// Context returns the root context any command should use.
+	Context() context.Context
 	// Clock returns a clock that provides access to the current time.
 	Clock() Clock
 	// GardenHomeDir returns the gardenctl home directory for the executing user.
@@ -50,6 +52,10 @@ type FactoryImpl struct {
 }
 
 var _ Factory = &FactoryImpl{}
+
+func (f *FactoryImpl) Context() context.Context {
+	return context.Background()
+}
 
 func (f *FactoryImpl) Manager() (target.Manager, error) {
 	cfg, err := config.LoadFromFile(f.ConfigFile)
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -23,7 +23,7 @@ import (
 type Factory interface {
 	// Clock returns a clock that provides access to the current time.
 	Clock() Clock
-	// GardenHomeDir returns the home directory for the executing user.
+	// GardenHomeDir returns the gardenctl home directory for the executing user.
 	GardenHomeDir() string
 	// Manager returns the target manager used to read and change the currently targeted system.
 	Manager() (target.Manager, error)
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -29,7 +29,7 @@ type Factory interface {
 	GardenHomeDir() string
 	// Manager returns the target manager used to read and change the currently targeted system.
 	Manager() (target.Manager, error)
-	// PublicIP returns the current host's public IP addresses. It's
+	// PublicIPs returns the current host's public IP addresses. It's
 	// recommended to provide a context with a timeout/deadline. The
 	// returned slice can contain IPv6, IPv4 or both, in no particular
 	// order.
--- a/internal/util/factory.go
+++ b/internal/util/factory.go
@@ -50,7 +50,12 @@ type FactoryImpl struct {
 	ConfigFile string
 
 	// TargetFile is the filename where the currently active target is located.
+	// This is only used if the TargetProvider property is nil.
 	TargetFile string
+
+	// TargetProvider can be used to completely override the provider. In this
+	// case, TargetFile is not used by the factory.
+	TargetProvider target.TargetProvider
 }
 
 var _ Factory = &FactoryImpl{}
@@ -65,7 +70,11 @@ func (f *FactoryImpl) Manager() (target.Manager, error) {
 		return nil, fmt.Errorf("failed to load config: %w", err)
 	}
 
-	targetProvider := target.NewFilesystemTargetProvider(f.TargetFile)
+	targetProvider := f.TargetProvider
+	if targetProvider == nil {
+		targetProvider = target.NewFilesystemTargetProvider(f.TargetFile)
+	}
+
 	kubeconfigCache := target.NewFilesystemKubeconfigCache(filepath.Join(f.GardenHomeDirectory, "cache", "kubeconfigs"))
 	clientProvider := target.NewClientProvider()
 
--- a/pkg/cmd/root/command.go
+++ b/pkg/cmd/root/command.go
@@ -34,7 +34,10 @@ const (
 )
 
 var (
-	factory = util.FactoryImpl{}
+	targetProvider = &DynamicTargetProvider{}
+	factory        = util.FactoryImpl{
+		TargetProvider: targetProvider,
+	}
 )
 
 // Execute adds all child commands to the root command and sets flags appropriately.
@@ -60,6 +63,12 @@ func Execute() {
 	// usage where the current user has no home directory (which might _just_ be
 	// the reason the user chose to specify an explicit config file).
 	rootCmd.PersistentFlags().StringVar(&factory.ConfigFile, "config", "", fmt.Sprintf("config file (default is $HOME/%s/%s.yaml)", gardenHomeFolder, configName))
+
+	// allow to temporarily re-target a different cluster
+	rootCmd.PersistentFlags().StringVar(&targetProvider.GardenNameFlag, "garden", "", "target the given garden cluster")
+	rootCmd.PersistentFlags().StringVar(&targetProvider.ProjectNameFlag, "project", "", "target the given project")
+	rootCmd.PersistentFlags().StringVar(&targetProvider.SeedNameFlag, "seed", "", "target the given seed cluster")
+	rootCmd.PersistentFlags().StringVar(&targetProvider.ShootNameFlag, "shoot", "", "target the given shoot cluster")
 
 	cobra.OnInitialize(initConfig)
 
@@ -117,6 +126,6 @@ func initConfig() {
 	}
 
 	factory.ConfigFile = viper.ConfigFileUsed()
-	factory.TargetFile = filepath.Join(home, targetFilename)
+	targetProvider.TargetFile = filepath.Join(home, targetFilename)
 	factory.GardenHomeDirectory = home
 }
--- a/pkg/cmd/root/command.go
+++ b/pkg/cmd/root/command.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 
 	"github.com/gardener/gardenctl-v2/internal/util"
+	"github.com/gardener/gardenctl-v2/pkg/cmd/ssh"
 	"github.com/gardener/gardenctl-v2/pkg/cmd/target"
 	"github.com/gardener/gardenctl-v2/pkg/cmd/version"
 
@@ -51,6 +52,7 @@ func Execute() {
 		SilenceUsage: true,
 	}
 
+	rootCmd.AddCommand(ssh.NewCommand(&factory, ssh.NewOptions(ioStreams)))
 	rootCmd.AddCommand(target.NewCommand(&factory, target.NewOptions(ioStreams)))
 	rootCmd.AddCommand(version.NewCommand(&factory, version.NewOptions(ioStreams)))
 
--- a/pkg/cmd/root/dynamic_target_provider.go
+++ b/pkg/cmd/root/dynamic_target_provider.go
@@ -1,0 +1,117 @@
+/*
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package root
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/gardener/gardenctl-v2/pkg/target"
+)
+
+// DynamicTargetProvider is a wrapper that combines the basic
+// file-based TargetProvider with CLI flags, to allow the user
+// to change the target for individual gardenctl commands
+// on-the-fly without changing the file on disk every time.
+//
+// If no CLI flags are given, this functions identical to the
+// regular TargetProvider from NewFilesystemTargetProvider().
+//
+// Otherwise, the flags are used to augment the existing target.
+// In this mode, changing the target via Write() is not allowed,
+// so that commands do not accidentally overwrite the target on
+// disk.
+type DynamicTargetProvider struct {
+	// TargetFile is the file where the target is read from / written to
+	// when no CLI flags override the targeting.
+	TargetFile string
+
+	GardenNameFlag  string
+	ProjectNameFlag string
+	SeedNameFlag    string
+	ShootNameFlag   string
+}
+
+// hasCLIFlags returns true if _any_ of the *Flag properties are not empty.
+func (p *DynamicTargetProvider) hasCLIFlags() bool {
+	return p.GardenNameFlag != "" || p.ProjectNameFlag != "" || p.SeedNameFlag != "" || p.ShootNameFlag != ""
+}
+
+// hasCompleteCLIFlags returns true if the set of given CLI flags is enough
+// to create a meaningful target. For example, if only the SeedNameFlag is
+// given, false is returned because for targeting a seed, the GardenNameFlag
+// must also be given. If ShootNameFlag and GardenNameFlag are set, false is
+// returned because either project or seed have to be given as well.
+func (p *DynamicTargetProvider) hasCompleteCLIFlags() bool {
+	// garden name is always required for a complete set of flags
+	if p.GardenNameFlag == "" {
+		return false
+	}
+
+	return target.NewTarget(p.GardenNameFlag, p.ProjectNameFlag, p.SeedNameFlag, p.ShootNameFlag).Validate() == nil
+}
+
+// Read returns the current target from the TargetFile if no CLI
+// flags were given, and tries to construct a meaningful target
+// otherwise.
+func (p *DynamicTargetProvider) Read() (target.Target, error) {
+	// user gave everything we needed
+	if p.hasCompleteCLIFlags() {
+		return target.NewTarget(p.GardenNameFlag, p.ProjectNameFlag, p.SeedNameFlag, p.ShootNameFlag), nil
+	}
+
+	// user didn't specify anything at all or _some_ flags; in both
+	// cases we need to read the current target from disk
+	current, err := target.NewFilesystemTargetProvider(p.TargetFile).Read()
+	if err != nil {
+		return nil, err
+	}
+
+	// user gave _some_ flags; we use those to override the current target
+	// (e.g. to quickly change a shoot while keeping garden/project names)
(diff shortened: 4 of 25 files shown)


pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+			}
+			if err := o.Validate(); err != nil {
+				return err
+			}
+
+			return runCommand(f, o)
+		},
+	}
+
+	cmd.Flags().BoolVar(&o.Interactive, "interactive", o.Interactive, "Open an SSH connection instead of just providing the bastion host.")
```
petersutter, line 67: add the information that a node name argument has to be provided (as you described it in the godoc on the `Options` struct for the `Interactive` flag)

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	cmd.Flags().StringArrayVar(&o.CIDRs, "cidr", nil, "CIDRs to allow access to the bastion host; if not given, the host's public IP is auto-detected.")
+	cmd.Flags().StringVar(&o.SSHPublicKeyFile, "public-key-file", "", "Path to the file that contains a public SSH key. If not given, a temporary keypair will be generated.")
```
petersutter, line 69: handle tilde :)

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	shoot, err := util.ShootForTarget(ctx, gardenClient, currentTarget)
+	if err != nil {
+		return err
+	}
+
+	// fetch the SSH key for the shoot nodes
+	nodePrivateKey, err := getShootNodePrivateKey(ctx, gardenClient, shoot)
```
petersutter, line 108: also fetch the old key in case there is one and handle fallback in case the ssh connection fails with the new key (e.g. `authorized_keys` file on the node was not yet updated).

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	bastionID, err := utils.GenerateRandomString(8)
+	if err != nil {
+		return fmt.Errorf("failed to create bastion name: %v", err)
+	}
+
+	bastionName := fmt.Sprintf("cli-%s", strings.ToLower(bastionID))
+	bastion := &operationsv1alpha1.Bastion{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      bastionName,
+			Namespace: "garden-dev", // shoot.Namespace,
```
petersutter, line 154: fix :)

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	fmt.Fprintf(o.IOStreams.Out, "Waiting up to %v for bastion to be ready…", o.WaitTimeout)
+
+	err = waitForBastion(ctx, o, gardenClient, bastion)
+
+	fmt.Fprintln(o.IOStreams.Out, "")
+
+	if err == wait.ErrWaitTimeout {
+		fmt.Fprintln(o.IOStreams.Out, "Timed out waiting for the bastion to be ready.")
+	} else if err != nil {
+		fmt.Fprintf(o.IOStreams.Out, "An error occured while waiting: %v", err)
```
petersutter, line 180:
```suggestion
		fmt.Fprintf(o.IOStreams.Out, "An error occurred while waiting: %v", err)
```

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	} else if ingress.Hostname != "" {
+		printAddr = ingress.Hostname
+	} else {
+		printAddr = ingress.IP
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Bastion host became available at %s.\n", printAddr)
+
+	// continuously keep the bastion alive by renewing its annotation
+	go keepBastionAlive(ctx, gardenClient, bastion, o.IOStreams.ErrOut)
```
petersutter, line 202: I'd expected this go-routine to be run before `waitForBastion`

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+	fmt.Fprintln(o.IOStreams.Out, "Exiting…")
+
+	// stop keeping the bastion alive
+	cancel()
+	<-ctx.Done()
+
+	return err
+}
+
+func preferedBastionAddress(bastion *operationsv1alpha1.Bastion) string {
```
petersutter, line 225: typo preferredBastionAddress

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+
+func waitForBastion(ctx context.Context, o *Options, gardenClient client.Client, bastion *operationsv1alpha1.Bastion) error {
+	return wait.Poll(5*time.Second, o.WaitTimeout, func() (bool, error) {
+		key := client.ObjectKeyFromObject(bastion)
+
+		if err := gardenClient.Get(ctx, key, bastion); err != nil {
+			return false, err
+		}
+
+		// TODO: update gardener dependency and use operationsv1alpha1.BastionReady const
```
petersutter, line 245: TODO (as reminder)

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+
+		if err := checkPortAvailable(checkCtx, preferedBastionAddress(bastion), SSHPort); err != nil {
+			return false, nil
+		}
+
+		return true, nil
+	})
+}
+
+// checkPortAvailable check whether the host with port is reachable within certain period of time
```
petersutter, line 263:
```suggestion
// checkPortAvailable checks whether the host with port is reachable within a certain period of time
```

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+				fmt.Fprintf(stderr, "Failed to keep bastion alive: %v\n", err)
+			}
+		}
+	}
+}
+
+func getShootNodePrivateKey(ctx context.Context, gardenClient client.Client, shoot *gardencorev1beta1.Shoot) ([]byte, error) {
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{
+		Name:      fmt.Sprintf("%s.ssh-keypair", shoot.Name),
```
petersutter, line 423: use `ShootProjectSecretSuffixSSHKeypair` constant?

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,460 @@
...
+		Namespace: shoot.Namespace,
+	}
+
+	if err := gardenClient.Get(ctx, key, secret); err != nil {
+		return nil, err
+	}
+
+	return secret.Data["id_rsa"], nil
```
petersutter, line 431: use `DataKeyRSAPrivateKey` constant?

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+		// run command
+		Expect(cmd.RunE(cmd, []string{"project", projectName})).To(Succeed())
+
+		currentTarget, err := targetProvider.Read()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(currentTarget.GardenName()).To(Equal(gardenName))
+		Expect(currentTarget.ProjectName()).To(Equal(projectName))
+	})
+})
```
petersutter, line 103: replace target tests with proper ssh command tests ;)

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
+	"github.com/gardener/gardener/pkg/utils"
+
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh"
+	cryptossh "golang.org/x/crypto/ssh"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+// Options is a struct to support target command
```
petersutter, line 35:
```suggestion
// Options is a struct to support ssh command
```

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+	base.Options
+
+	// Interactive can be used to toggle between gardenctl just
+	// providing the bastion host while keeping it alive (non-interactive),
+	// or gardenctl opening the SSH connection itself (interactive). For
+	// interactive mode, a NodeName must be specified as well.
+	Interactive bool
+
+	// NodeName is the name of the Shoot cluster node that the user wants to
+	// connect to. If this is left empty, grdenctl will only establish the
```
petersutter, line 46:
```suggestion
	// connect to. If this is left empty, gardenctl will only establish the
```

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+	// bastion host, but leave it up to the user to SSH themselves.
```
petersutter, line 47: Something like this could also be added to the command help text

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+	NodeName string
+
+	// CIDRs is a list of IP address ranges to allowed for accessing the
```
petersutter, line 50: >CIDRs is a list of IP address ranges to allowed for accessing the created Bastion host

the "to" does not fit into the sentence or something is missing?
`CIDRs is a list of IP address ranges allowed for accessing the created Bastion host` or
`CIDRs is a list of IP address ranges to be / that should be allowed for accessing the created Bastion host`

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+		},
+		Interactive: true,
+		WaitTimeout: 10 * time.Minute,
+	}
+}
+
+// Complete adapts from the command line args to the data required.
+func (o *Options) Complete(f util.Factory, cmd *cobra.Command, args []string, stdout io.Writer) error {
+	if len(o.CIDRs) == 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
petersutter, line 88: 5 seconds timeout? why so short? I'd go with the default 60seconds timeout

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+		defer cancel()
+
+		publicIP, err := f.PublicIP(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to determine your system's public IP address: %w", err)
```
petersutter, line 93: the help text of the cidr flag is

> CIDRs to allow access to the bastion host; if not given, the host's public IP is auto-detected.

To better relate the `cidr` flag to the error message either change the help text to
`CIDRs to allow access to the bastion host; if not given, your system's public IP is auto-detected.`

or adapt this error message accordingly and use "host" instead of "system"

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+		}
+
+		fmt.Fprintf(stdout, "Auto-detected public IP as %s\n", publicIP)
```
petersutter, line 96: I'd prefer to print out the cidr, as this is also the format that we accept as flag

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,228 @@
...
+
+		o.CIDRs = append(o.CIDRs, ipToCIDR(publicIP))
+	}
+
+	if len(o.SSHPublicKeyFile) == 0 {
+		privateKeyFile, publicKeyFile, err := createSSHKeypair("", "")
```
petersutter, line 102: cleanup keys on gardenctl exit or wasted effort?

pkg/cmd/ssh/options_test.go
```diff
@@ -0,0 +1,37 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		streams, _, _, _ := genericclioptions.NewTestIOStreams()
+		o := NewOptions(streams)
+		o.CIDRs = []string{"8.8.8.8/32"}
+		o.SSHPublicKeyFile = tmpFile.Name()
+
+		Expect(o.Validate()).To(Succeed())
+	})
+})
```
petersutter, line 37: also add some validation error tests :)

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,467 @@
...
+	if node != nil && o.Interactive {
+		err = remoteShell(ctx, o, bastion, nodePrivateKeyFile, node)
+	} else {
+		err = waitForSignal(o, bastion, nodePrivateKeyFile, node)
+	}
+
+	fmt.Fprintln(o.IOStreams.Out, "Exiting…")
+
+	if !o.KeepBastion {
+		if err := gardenClient.Delete(ctx, bastion); err != nil {
```
petersutter, line 213:
```suggestion
		fmt.Fprintf(o.IOStreams.Out, "Deleting bastion %s…", bastion.Name)

		if err := gardenClient.Delete(ctx, bastion); err != nil {
```
add logs statement

internal/util/factory.go
```diff
@@ -56,10 +64,38 @@ func (f *FactoryImpl) Manager() (target.Manager, error) {
...
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	ip, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", err
+	}
+
+	ipAddress := strings.TrimSpace(string(ip))
```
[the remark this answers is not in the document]
xrstf: And why? What does JSON offer in this case, compared to plaintext?

pkg/cmd/ssh/command.go
xrstf: Your shell already takes care of that.

pkg/cmd/ssh/command.go
xrstf: Do we really want to implement something here that doesn't exist yet in Gardener?

pkg/cmd/ssh/command.go
xrstf: We cannot use that package because once we import it, we get

> ../../../../pkg/mod/github.com/gardener/gardener@v1.23.0/pkg/client/kubernetes/runtime_client.go:84:9: undefined: manager.NewClientBuilder

As mentioned in another comment, we probably need to wait for ctrlruntime 0.9 support in Gardener.

I can leave a TODO comment though.

pkg/cmd/ssh/command.go
xrstf: Same problem as for `ShootProjectSecretSuffixSSHKeypair` :-/

pkg/cmd/ssh/options.go
xrstf: We wanted to leave the keys, because otherwise keeping the bastion would be pointless.

pkg/cmd/ssh/options.go
```diff
@@ -0,0 +1,234 @@
...
+
+// NewOptions returns initialized Options
+func NewOptions(ioStreams genericclioptions.IOStreams) *Options {
+	return &Options{
+		Options: base.Options{
+			IOStreams: ioStreams,
+		},
+		Interactive: true,
+		WaitTimeout: 10 * time.Minute,
+		KeepBastion: false,
```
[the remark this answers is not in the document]
xrstf: Because this way _everyone_ can easily get a hold of the default config, and it's not just buried in the CLI setup code. In my mind, the "ownership" of the default values belongs to the options struct, not one of its users (the cobra command).

internal/util/factory.go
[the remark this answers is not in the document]
petersutter: I do not insist on changing it. This is just my personal preference to work with structured input but as you validate the response by trying to parse it as ip afterwards it does not really matter in this case.

pkg/cmd/ssh/command.go
petersutter: then let's do it in a follow up pr

pkg/cmd/ssh/command.go
petersutter: ok then add a TODO comment

pkg/cmd/ssh/command.go
petersutter: yes, just leave a TODO comment

pkg/cmd/ssh/options.go
petersutter: with `keep-bastion` set to false they wouldn't be needed ;)
anyhow, maybe not worth the effort

pkg/cmd/ssh/options.go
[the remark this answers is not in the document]
petersutter: allright

pkg/cmd/ssh/options.go
petersutter: Yes could be an option

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,541 @@
...
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Creating bastion %s…", bastion.Name)
+
+	if err := gardenClient.Create(ctx, bastion); err != nil {
+		return fmt.Errorf("failed to create bastion: %v", err)
+	}
+
+	// continuously keep the bastion alive by renewing its annotation
+	go keepBastionAlive(ctx, gardenClient, bastion, o.IOStreams.ErrOut)
```
petersutter, line 186: do not reuse the bastion resource in the go-routine, as this could lead to unexpected side-effects

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,541 @@
...
+	if err == wait.ErrWaitTimeout {
+		fmt.Fprintln(o.IOStreams.Out, "Timed out waiting for the bastion to be ready.")
+	} else if err != nil {
+		fmt.Fprintf(o.IOStreams.Out, "An error occurred while waiting: %v", err)
+	}
+
+	if err != nil {
+		// actual error has already been printed
+		return errors.New("precondition failed")
+	}
```
petersutter, line 203: I guess we also need to run the cleanup, depending on o.KeepBastion, or not?

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,541 @@
...
+
+	// stop keeping the bastion alive
+	cancel()
+	<-ctx.Done()
+
+	return err
+}
+
+func getNodeNamesFromShoot(f util.Factory, o *Options, prefix string) ([]string, error) {
```
petersutter, line 246: unused options param?

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,541 @@
...
+	for _, k := range []corev1.NodeAddressType{corev1.NodeInternalIP, corev1.NodeInternalDNS, corev1.NodeExternalIP, corev1.NodeExternalDNS} {
+		if addr := addresses[k]; addr != "" {
+			return addr, nil
+		}
+	}
+
+	return "", errors.New("node has no internal or external names")
+}
+
+func getShootNodes(ctx context.Context, shootClient client.Client) ([]corev1.Node, error) {
```
petersutter, line 534:
```go
func getNodes(ctx context.Context, client client.Client) ([]corev1.Node, error) {
```

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,541 @@
...
+		Spec: operationsv1alpha1.BastionSpec{
+			ShootRef: corev1.LocalObjectReference{
+				Name: shoot.Name,
+			},
+			SSHPublicKey: strings.TrimSpace(string(sshPublicKey)),
+			Ingress:      policies,
+		},
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Creating bastion %s…", bastion.Name)
```
petersutter, line 179:
```bash
gardenctl ssh shoot--myproject--mynode
Auto-detected your system's CIDR as <redacted>/64
Creating bastion cli-uhgvgvnb…Waiting up to 10m0s for bastion to be ready…
Timed out waiting for the bastion to be ready.
Error: precondition failed
```
```bash
gardenctl-v2 ssh
Auto-detected your system's CIDR as <redacted>/64
Creating bastion cli-kpvvyqof…Waiting up to 10m0s for bastion to be ready…
An error occurred while waiting: bastions.operations.gardener.cloud "cli-kpvvyqof" not foundError: precondition failed
```
See the missing linebreaks above. And why do you use `fmt.Fprintln(o.IOStreams.Out, "")` instead of `\n` at the end of the previous `Fprintf` message?

petersutter, line 179: One additional thing: in my case the bastion creation failed. Unfortunately there is little to no feedback on the current status. Maybe it would help when we show the current state of the BastionReady condition? WDYT?

xrstf: Because I like it when a status messag like "creating something..." keeps the cursor on the same line. But it's not important, I changed it now.

xrstf: I've adjusted the wait loop to print the last error before the timeout. WDYT?

petersutter:
```bash
gardenctl-v2 ssh ip-10-<redacted>.compute.internal
Auto-detected your system's CIDR as <redacted>/64
Creating bastion cli-j79p1oum…
Waiting up to 10m0s for bastion to be ready…
An error occurred while waiting: timed out waiting for the bastion to become ready: dial tcp <redacted>:22: i/o timeoutDeleting bastion cli-j79p1oum…
Error: precondition failed
```
Also here some missing linebreaks ;)

It's better, but currently, if it doesn't work, you just see `Waiting up to 10m0s for bastion to be ready…` for 10 minutes straight until it times out and you see the last error.
Wouldn't it be helpful for the user to see what's currently going on / show more details (maybe with verbose option)? E.g. when already some minutes passed by I wonder what the status is, do we already have an ip for the jump host ? Is gardenctl trying to connect but it fails? WDYT?

pkg/cmd/ssh/command_test.go
petersutter: could it help if you start a local server that acts as the bastion host?

xrstf: I think at that point we firmly leave the territory of unit tests and are in full-on integration/e2e test mode.

xrstf: Let's meet in the middle and do some whitebox tests. I thought about blackbox tests, but that would require us to introduce a lot of very artifical looking interfaces to the factory that will most likely only be ever used by the SSH command.

I've added some tests, doing the minimum amount of mocking required to run most of the actual codebase.

petersutter:
when I run `gardenctl-v2 ssh <redacted>.internal` and press `ctrl + c`  after the bastion was created the signal is not handled and it does not get cleaned up

internal/fake/factory.go
```diff
@@ -42,13 +47,13 @@ func NewFakeManagerFactory(manager target.Manager) util.Factory {
 	}
 }
 
-func NewFakeFactory(cfg *config.Config, clock util.Clock, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) util.Factory {
+func NewFakeFactory(cfg *config.Config, clock util.Clock, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) *Factory {
```
petersutter, line 50: I guess the other `NewFakeManagerFactory` function is not used anymore

pkg/cmd/ssh/command.go
```diff
@@ -194,6 +280,11 @@ func runCommand(f util.Factory, o *Options) error {
 				_ = os.Remove(o.SSHPublicKeyFile)
 				_ = os.Remove(o.SSHPrivateKeyFile)
 			}
+
+			// though technically not used _on_ the bastion itself, without
+			// this file remaining, the user would not be able to use the SSH
+			// command we provided to connect to the shoot nodes
+			defer os.Remove(nodePrivateKeyFile)
```
petersutter, line 287: does it make sense to print out the left-overs that are not cleaned up in case keepBastion is true, so that the user has a chance to delete / reuse them? WDYT

petersutter (commented):
command_test.go still to be reviewed..

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,625 @@
...
+	bastionNameProvider = func() (string, error) {
+		bastionID, err := utils.GenerateRandomString(8)
+		if err != nil {
+			return "", err
+		}
+
+		return fmt.Sprintf("cli-%s", strings.ToLower(bastionID)), nil
+	}
+
+	// createSignalChannel returns a chanel which receives OS signals.
```
petersutter, line 115:
```suggestion
	// createSignalChannel returns a channel which receives OS signals.
```

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,404 @@
...
+
+		// let the magic happen
+		Expect(cmd.RunE(cmd, nil)).To(Succeed())
+
+		// assert the output
+		Expect(out.String()).To(ContainSubstring(bastionName))
+		Expect(out.String()).To(ContainSubstring(bastionHostname))
+		Expect(out.String()).To(ContainSubstring(bastionIP))
+
+		// asser that the bastion has been cleaned up
```
petersutter, line 239:
```suggestion
		// assert that the bastion has been cleaned up
```

pkg/cmd/ssh/command_test.go
```diff
@@ -4,100 +4,401 @@ SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener con
...
+		execCommand = func(ctx context.Context, command string, args []string, o *Options) error {
+			executedCommands++
+			return nil
+		}
+
+		// let the magic happen
+		Expect(cmd.RunE(cmd, []string{testNode.Name})).To(Succeed())
+
+		// assert output
+		Expect(executedCommands).To(Equal(1))
```
petersutter, line 321: it would be nice to also validate the command / args

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,404 @@
...
+			factory := internalfake.NewFakeFactory(cfg, nil, clientProvider, kubeconfigCache, targetProvider)
+			factory.ContextImpl = ctx
+
+			options := NewOptions(streams)
+			cmd := NewCommand(factory, options)
+
+			// let the magic happen; should find "monitoring" node based on this prefix
+			suggestions, directive := cmd.ValidArgsFunction(cmd, nil, "mon")
+			Expect(directive).To(Equal(cobra.ShellCompDirectiveNoFileComp))
+			Expect(suggestions).To(HaveLen(1))
```
petersutter, line 401:
```suggestion
			Expect(suggestions).To(HaveLen(1))
			Expect(suggestions).To(Equal([]string{"monitoring"}))
```

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,404 @@
...
+		execCommand = func(ctx context.Context, command string, args []string, o *Options) error {
+			executedCommands++
+			return nil
+		}
+
+		// let the magic happen
+		Expect(cmd.RunE(cmd, []string{testNode.Name})).To(Succeed())
+
+		// assert output
+		Expect(executedCommands).To(Equal(1))
```
petersutter, line 321: same applies here as well

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,404 @@
...
+
+		// assert that no temporary SSH keypair remained on disk
+		_, err := os.Stat(options.SSHPublicKeyFile)
+		Expect(err).To(HaveOccurred())
+
+		_, err = os.Stat(options.SSHPrivateKeyFile)
+		Expect(err).To(HaveOccurred())
+	})
+
+	It("should connect to a given node", func() {
```
petersutter, line 253: would be nice to  have a test for the keep alive annotation

pkg/cmd/ssh/command.go
```diff
@@ -61,6 +114,7 @@ func NewCommand(f util.Factory, o *Options) *cobra.Command {
 			nodeNames, err := getNodeNamesFromShoot(f, toComplete)
 			if err != nil {
 				fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
+				fmt.Println(err.Error())
```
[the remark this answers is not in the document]
xrstf: Correct!

pkg/cmd/ssh/command.go
xrstf: Can do

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,636 @@
...
+		},
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Creating bastion %s…\n", bastion.Name)
+
+	if err := gardenClient.Create(ctx, bastion); err != nil {
+		return fmt.Errorf("failed to create bastion: %v", err)
+	}
+
+	defer func() {
```
petersutter, line 273: extract function so that the `runCommand` get's a bit shorter

pkg/cmd/ssh/command.go
petersutter: sorry, what I meant was to handle the `else`-case
```go
if !o.KeepBastion {
    // ..
} else {
  // print out left-overs that are not cleaned
}

File: pkg/cmd/ssh/command_test.go [hunk 40]
```diff
@@ -337,6 +366,67 @@ var _ = Describe("Command", func() {
...
+		}
+
+		// let the magic happen
+		Expect(cmd.RunE(cmd, nil)).To(Succeed())
+
+		key := types.NamespacedName{Name: bastionName, Namespace: *testProject.Spec.Namespace}
+		bastion := &operationsv1alpha1.Bastion{}
+
+		Expect(gardenClient.Get(ctx, key, bastion)).To(Succeed())
+		Expect(bastion.Annotations).To(HaveKeyWithValue(v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationKeepalive))
```
petersutter, line 427: Use `Eventually` instead of `Expect`, see https://onsi.github.io/gomega/#making-asynchronous-assertions.
You can then call `close(signalChan)` afterwards and only have something like this as `createSignalChannel`
```go
signalChan := make(chan os.Signal, 1)
createSignalChannel = func() <-chan os.Signal {
	return signalChan
}
```

Maybe it also makes sense to shorten the `keepAliveInterval` to some milliseconds so the test runs even quicker

xrstf: I'm not sure using `Eventually` would work here. If nobody closes the signal channel, the `RunE()` will wait and eventually time out, returning an error.

petersutter: ok then some more changes are necessary. Here I tried it out quick and dirty:

```go
		signalChan := make(chan os.Signal, 1)
		createSignalChannel = func() <-chan os.Signal {
			return signalChan
		}

		key := types.NamespacedName{Name: bastionName, Namespace: *testProject.Spec.Namespace}

		go func() {
			defer GinkgoRecover()

			Eventually(func() bool {
				bastion := &operationsv1alpha1.Bastion{}
				if err := gardenClient.Get(ctx, key, bastion); errors.IsNotFound(err) {
					return false
				}

				return bastion.Annotations != nil && bastion.Annotations[v1beta1constants.GardenerOperation] == v1beta1constants.GardenerOperationKeepalive
			}).Should(BeTrue())
			close(signalChan)
		}()

		// let the magic happen
		Expect(cmd.RunE(cmd, nil)).To(Succeed())

		// Double check that the annotation was really set
		bastion := &operationsv1alpha1.Bastion{}
		Expect(gardenClient.Get(ctx, key, bastion)).To(Succeed())
		Expect(bastion.Annotations).To(HaveKeyWithValue(v1beta1constants.GardenerOperation, v1beta1constants.GardenerOperationKeepalive))
```

It has the advantage that the test terminates as soon as the annotation is set - of course depending on the (default) interval of `Eventually`.
It also makes it more resilient in case it takes a bit longer than anticipated (assuming the timeout on `Eventually` is high enough.

pkg/cmd/ssh/command.go
```diff
@@ -339,6 +314,41 @@ func runCommand(f util.Factory, o *Options) error {
...
+			fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete bastion: %v", err)
+		}
+
+		if o.generatedSSHKeys {
+			if err := os.Remove(o.SSHPublicKeyFile); err != nil {
+				fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete SSH public key file %q: %v", o.SSHPublicKeyFile, err)
+			}
+
+			if err := os.Remove(o.SSHPrivateKeyFile); err != nil {
+				fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete SSH private key file %q: %v", o.SSHPrivateKeyFile, err)
```
petersutter, line 331:
```suggestion
				fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete SSH public key file %q: %v\n", o.SSHPublicKeyFile, err)
			}

			if err := os.Remove(o.SSHPrivateKeyFile); err != nil {
				fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete SSH private key file %q: %v\n", o.SSHPrivateKeyFile, err)
```

newline missing?

pkg/cmd/ssh/command_test.go
```diff
@@ -0,0 +1,481 @@
...
+
+			patch := client.MergeFrom(bastion.DeepCopy())
+			patcher(&bastion.Status)
+
+			Expect(gardenClient.Status().Patch(ctx, bastion, patch)).To(Succeed())
+
+			return
+		}
+	}
+}
```
petersutter, line 70:
```suggestion
func waitForBastionThenPatchStatus(ctx context.Context, gardenClient client.Client, bastionName string, namespace string, patcher bastionStatusPatch) {
	defer GinkgoRecover()

	ticker := time.NewTicker(1 * time.Second)
	defer ticker.Stop()

	for {
		select {
		case <-ctx.Done():
			return

		case <-ticker.C:
			key := types.NamespacedName{Name: bastionName, Namespace: namespace}

			Eventually(func() error {
				bastion := &operationsv1alpha1.Bastion{}

				if err := gardenClient.Get(ctx, key, bastion); err != nil {
					return err
				}

				patch := client.MergeFrom(bastion.DeepCopy())
				patcher(&bastion.Status)

				return gardenClient.Status().Patch(ctx, bastion, patch)
			}).Should(Succeed())

			return
		}
	}
}
```

We have to add `defer GinkgoRecover()` as this function is always run in a go routine.

I also had a case where the status patch failed with an error that the bastion was modified. That's why I wrapped it with `Eventually`

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,656 @@
...
+	return err
+}
+
+func cleanup(ctx context.Context, o *Options, gardenClient client.Client, bastion *operationsv1alpha1.Bastion, nodePrivateKeyFile string) {
+	if !o.KeepBastion {
+		fmt.Fprintf(o.IOStreams.Out, "Deleting bastion %s…\n", bastion.Name)
+
+		if err := gardenClient.Delete(ctx, bastion); err != nil {
+			fmt.Fprintf(o.IOStreams.ErrOut, "Failed to delete bastion: %v", err)
+		}
```
petersutter, line 331: ignore not found error

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,801 @@
...
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{
+		// TODO: use ShootProjectSecretSuffixSSHKeypair once Gardener supports ctrl-runtime 0.9
+		Name:      fmt.Sprintf("%s.ssh-keypair", shoot.Name),
+		Namespace: shoot.Namespace,
+	}
+
+	if err := gardenClient.Get(ctx, key, secret); err != nil {
+		return nil, err
+	}
```
petersutter, line 736: there might be a time window where this secret is not there, see https://github.com/gardener/gardener/blob/f54564c83d8dee91eb46bc726119a8d72d578810/pkg/operation/botanist/secrets.go#L287-L289

I think in this case we could also ignore not found errors. 

However if `keys` slice is empty (both secrets were not found) it should be considered as error, of course

pkg/cmd/root/command.go
```diff
@@ -61,6 +64,12 @@ func Execute() {
 	// the reason the user chose to specify an explicit config file).
 	rootCmd.PersistentFlags().StringVar(&factory.ConfigFile, "config", "", fmt.Sprintf("config file (default is $HOME/%s/%s.yaml)", gardenHomeFolder, configName))
 
+	// allow to temporarily re-target a different cluster
```
petersutter, line 67: expect when using on the `target` command, or not?

pkg/cmd/ssh/command.go
xrstf: I don't quite understand. If the "primary" SSH key isn't there, the secondary (".old") will also not be there. So if we ignore the NotFound error on the primary, the secondary will also fail and we will run into the "slice is empty" situation, i.e. an error.

Or are you referring to a different timeframe than shortly after a Shoot has been created?

petersutter: during `rotateSSHKeypairSecrets` (see link above) the current secret get's deleted and recreated later, however at that point in time the old key should exist

```go
	// delete the current key so it gets re-generated later
	if err := b.K8sSeedClient.Client().Delete(ctx, currentSecret); client.IgnoreNotFound(err) != nil {
		return fmt.Errorf("failed to delete current SSH keypair: %v", err)
	}
```

pkg/cmd/ssh/command.go
```diff
@@ -0,0 +1,801 @@
...
+		Short: "Establish an SSH connection to a Shoot cluster's node",
+		Args:  cobra.MaximumNArgs(1),
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			if len(args) != 0 {
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			}
+
+			nodeNames, err := getNodeNamesFromShoot(f, toComplete)
+			if err != nil {
+				fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
```
petersutter, line 156: for debugging purposes, cobra suggests to use https://github.com/spf13/cobra/blob/master/shell_completions.md#dynamic-completion-of-nouns

```go
// Prints to the completion script debug file (if BASH_COMP_DEBUG_FILE
// is set to a file path) and optionally prints to stderr.
cobra.CompDebug(msg string, printToStdErr bool) {
cobra.CompDebugln(msg string, printToStdErr bool)

// Prints to the completion script debug file (if BASH_COMP_DEBUG_FILE
// is set to a file path) and to stderr.
cobra.CompError(msg string)
cobra.CompErrorln(msg string)
```

however you can't pass your own IOStream unfortunately.

xrstf: So what sacrifice do we want to make here? ^^

petersutter: I give you the honour to choose :) I do not have a strong opinion here

pkg/cmd/root/dynamic_target_provider.go
```diff
@@ -0,0 +1,116 @@
...
+// on-the-fly without changing the file on disk every time.
+//
+// If no CLI flags are given, this functions identical to the
+// regular TargetProvider from NewFilesystemTargetProvider().
+//
+// Otherwise, the flags are used to augment the existing target.
+// In this mode, changing the target via Write() is not allowed,
+// so that commands do not accidentally overwrite the target on
+// disk.
+type DynamicTargetProvider struct {
```
petersutter, line 28: DynamicTargetProvider implements `target.TargetProvider` right?
Why have you chosen to have `dynamic_target_provider.go` under the `root` package and not under the `target` package?

xrstf: Because the DTP only makes sense in the context of the CLI flags, so I did not want to introduce a cobra dependency in the `target` package.

pkg/cmd/ssh/command.go
xrstf: Then I would simply keep it as it. The "convenience" of replacing a single `fmt.Println()` call with something else is not worth losing the abstraction that is IOStreams.

petersutter: Fine 👍

petersutter:
the test failed with because of a race condition

petersutter (approved):
/lgtm

thank you very much for your efforts! Really nice PR!

PR #3, opened 2021-05-20
propose structure for commands

xrstf:
**What this PR does / why we need it**:
This PR demonstrates a possible structure for this repository, using the somewhat simple `version` command as an example.

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature user
NONE
```

PR #5, opened 2021-05-25
add targeting system and target command

xrstf:
This PR introduces the "targeting" system from gardenctl v1, just slighly refactored. Targeting is used to free the user from having to specify the garden, seed and/or shoot name for most commands (e.g. free them from doing something like `gardenctl ssh -g garden1 -s seed1 --ss shoot1 nodeNameXY`). To achieve this, the user can use the `target` command and incrementally refine the goal for the upcoming commands, e.g.

```bash
gardenctl target garden mygarden
gardenctl target seed aws-seed
gardenctl target shoot my-shoot
gardenctl ssh nodeNameXY
```

The previous implementation chose to work with "stack"-like approach, however the new implementation uses a dedicated map instead. The code is also less generic, which seems like a downside, but if you look carefully at the v1 code, the generic parts were drowned in super specific hacks and exceptions and lots and lots of assumptions. That made the old code hard to read and understand:

```go
// getKubeConfigOfClusterType return config of specified type
func getKubeConfigOfClusterType(clusterType TargetKind) (pathToKubeconfig string) {
	var target Target
	ReadTarget(pathTarget, &target)
	gardenName := target.Stack()[0].Name
	switch clusterType {
	case TargetKindGarden:
		pathToKubeconfig = TidyKubeconfigWithHomeDir(getGardenKubeConfig())
	case TargetKindSeed:
		if target.Target[1].Kind == "seed" {
			pathToKubeconfig = filepath.Join(pathGardenHome, "cache", gardenName, "seeds", target.Target[1].Name, "kubeconfig.yaml")
		} else {
			pathToKubeconfig = filepath.Join(pathGardenHome, "cache", gardenName, "seeds", getSeedForProject(target.Target[2].Name), "kubeconfig.yaml")
		}
	case TargetKindShoot:
		if target.Target[1].Kind == "seed" {
			pathToKubeconfig = filepath.Join(pathGardenHome, "cache", gardenName, "seeds", getSeedForProject(target.Target[2].Name), target.Target[2].Name, "kubeconfig.yaml")
		} else if target.Target[1].Kind == "project" {
			pathToKubeconfig = filepath.Join(pathGardenHome, "cache", gardenName, "projects", target.Target[1].Name, target.Target[2].Name, "kubeconfig.yaml")
		}
	}
	return pathToKubeconfig
}
```

The new v2 code makes no assumptions like that and simply has dedicated functions and fields for garden, seed, project and shoot.

Likewise, the new code separates the data and logic, with the `Target` interface only providing _read_ access to target information, and the more complex logic being wrapped up in the `Manager` (so it's `manager.GardenClient(target)` instead of `target.K8sClientForTarget(TargetKindGarden)`). This is because we surely will rework the targeting in the future and this separation makes it easier to exchange the implementation.

I chose to provide controller-runtime clients, because I like their interface much better than the Kuberntes clientsets. One downside so far is that namespace targeting doesn't work yet, so targeting a project is currently pointless and returns a client to the garden cluster.

Fixes #

**Release note**:

```feature user
Add `target` command
```


pkg/cmd/target/options.go
```diff
@@ -0,0 +1,89 @@
...
+func (o *Options) Complete(f util.Factory, cmd *cobra.Command, args []string) error {
+	if len(args) != 2 {
+		return errors.New("expected exactly 2 arguments")
+	}
+
+	o.Kind = TargetKind(strings.TrimSpace(args[0]))
+	o.TargetName = strings.TrimSpace(args[1])
+
+	return nil
+}
```
petersutter, line 67: have you thought about using hierarchical target commands instead?
something like this structure
```yaml
cmds:
- target:
    # flags: # need to think about that
    # - garden
    # - seed
    # - project
    # - shoot
    cmds:
    - garden:
        flags: ~
    - seed:
        flags:
        - garden
    - project:
        flags: 
        - garden
    - shoot:
        flags:
        - garden
        - seed
        - project
- ssh: 
# [...]
```

xrstf: Seeing how simple the command logic is now, I don't think it would offer many benefits when we split it into subcommands. I was taking a quick glance at cobra's auto completion magic and it seemed like we should be able to offer nice completion even if it's a single command. But that was just a very quick glance ^^

pkg/target/config.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"gopkg.in/yaml.v3"
+)
+
+type Config struct {
+	Gardens []Garden `yaml:"gardens"`
```
petersutter, line 16: could you add a config option so that kubeconfigs are cloned when targeting?

See also https://github.com/gardener/gardenctl/issues/499#issuecomment-749380248 "Use case b"
> I'm targeting some control plane and now want to analyze another control plane on the same seed side-by-side in multiple panes.
What I currently do for this is "cloning" my kubeconfig, which allows me to use kubens to have to panes targeting the same cluster but different namespaces.

> [...]
Therefore, I am clearly for option #3. Every time you target something or retarget it, it's another "instance" of the kubeconfig.

pkg/target/kubeconfig_cache_test.go
```diff
@@ -0,0 +1,96 @@
...
+		readBack, err := cache.Read(targetA)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(readBack).To(Equal(dataA))
+
+		readBack, err = cache.Read(targetB)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(readBack).To(Equal(dataB))
+	})
+
+	It("should separate pased on the path", func() {
```
petersutter, line 79:
```suggestion
	It("should separate based on the path", func() {
```

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	target, err := m.targetProvider.Read()
+	if err != nil {
+		return err
+	}
+
+	// this is horrible cheating
+	impl, ok := target.(*targetImpl)
+	if !ok {
+		return errors.New("target must be using targetImpl as its underlying type")
+	}
```
petersutter, line 383: you need to have setters on the target interface instead, right?

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+			return fmt.Errorf("failed to validate seed: %w", err)
+		}
+
+		t.Seed = seedName
+		t.Project = ""
+		t.Shoot = ""
+
+		if err := m.updateSeedKubeconfig(ctx, gardenClient, t, seed); err != nil {
+			return fmt.Errorf("failed to fetch kubeconfig for seed cluster: %w", err)
+		}
```
petersutter, line 136: this shouldn't be part of the patch func

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	seed := &gardencorev1beta1.Seed{}
+	key := types.NamespacedName{Name: seedName}
+
+	if err := gardenClient.Get(ctx, key, seed); err != nil {
+		return nil, err
+	}
+
+	if seed.Spec.SecretRef == nil {
+		return nil, errors.New("spec.SecretRef is missing in this seed, seed not reachable")
+	}
```
petersutter, line 164: this check does not fit to the function scope / function name

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+		return nil
+	})
+}
+
+// updateShootKubeconfig must be called with *either* project or seed, but never both
+// and never neither.
+func (m *managerImpl) updateShootKubeconfig(
+	ctx context.Context,
+	gardenClient client.Client,
+	project *gardencorev1beta1.Project,
```
petersutter, line 216: generally we could just accept the namespace here instead of the project, or not?

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	seed *gardencorev1beta1.Seed,
+	t Target,
+) error {
+	shoot := &gardencorev1beta1.Shoot{}
+
+	// If a shoot is targeted via a project, we fetch the project and find
+	// the shoot by listing all shoots in the project's spec.namespace.
```
petersutter, line 223: This comment is not correct. If we have the project / namespace, we can just get the shoot (as you do below). No need to list anything

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+			return fmt.Errorf("invalid shoot %q: not found on seed %q", t.ShootName(), seed.Name)
+		}
+
+		if len(matchingShoots) > 1 {
+			return fmt.Errorf("there are multiple shoots named %q on this garden, please target via project to make your choice unambiguous", t.ShootName())
+		}
+
+		shoot = matchingShoots[0]
+	} else {
+		return errors.New("neither project nor seed were provided to filter the shoot by")
```
petersutter, line 262: well basically, similar as when a seed is targeted, we can go over all the shoots to find it

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	return seed, nil
+}
+
+func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
+	return m.patchTarget(func(t *targetImpl) error {
+		if t.Garden == "" {
+			return ErrNoGardenTargeted
+		}
+
+		if t.Project == "" && t.Seed == "" {
```
petersutter, line 175: as written here https://github.com/gardener/gardenctl-v2/pull/5/files#r640661421 it shouldn't be a hard requirement that a project or seed is targeted. We could figure the out the project when searching for the shoot across all namespaces

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+	kubeconfig, err := m.kubeconfigCache.Read(t)
+	if err != nil {
+		return nil, err
+	}
+
+	return m.clientProvider.FromBytes(kubeconfig)
+}
+
+func (m *managerImpl) ShootNamespaceClient(t Target) (client.Client, error) {
+	return m.SeedClient(t) // TODO: pre-configure the client to use a certain namespace?
```
petersutter, line 344: yes :)

internal/fake/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	TargetFile string
+}
+
+func NewFakeManagerFactory(manager target.Manager) util.Factory {
+	return &Factory{
+		ManagerImpl: manager,
+	}
+}
+
+func NewFakeFactory(config *target.Config, clientProvider target.ClientProvider, kubeconfigCache target.KubeconfigCache, targetProvider target.TargetProvider) util.Factory {
```
petersutter, line 50: what about the clock?

internal/fake/factory.go
```diff
@@ -0,0 +1,93 @@
...
+	// the garden home.
+	GardenHomeDirectory string
+
+	// ConfigFile is the location of the gardenctlv2 configuration file.
+	// This can be overriden via a CLI flag and defaults to ~/.garden/gardenctlv2.yaml
+	// if empty.
+	ConfigFile string
+
+	// TargetFile is the filename where the currently active target is located.
+	TargetFile string
```
petersutter, line 41: is this needed for the fake?

pkg/cmd/target/options.go
petersutter: Still, I think I would prefer to have it split up into subcommands. You get rid of the switch, don't need to mess around with custom completion logic and parts of the options validation is also obsolet as it is handled by cobra. Let's also talk how it would look like if we evaluate also the flags, e.g. `g target shoot -g canary -p core funny` once you are back

pkg/target/config.go
```diff
@@ -0,0 +1,59 @@
+/*
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+package target
```
petersutter, line 6: shouldn't config.go be under a more common package as it applies also for the other commands

pkg/target/manager.go
xrstf: I want to prevent users from mucking with the target by accident. They were supposed to get some immutable data structure and were supposed to always refer back to the manager if they want another target. That seems to be a worthwhile design decision to keep the interface and teeth grindingly accept this ...cheat.

pkg/target/manager.go
xrstf: Why is that? If storing the kubeconfig locally fails, I do not want a "broken" target on disk.

pkg/target/manager.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// If a shoot is targeted via a project, we fetch the project and find
+	// the shoot by listing all shoots in the project's spec.namespace.
+	// If the target uses a seed, _all_ shoots in the garden are filtered
+	// for shoots with matching seed and name. It's an error if no or multiple
+	// matching shoots are found.
+
+	if project != nil {
+		// fetch shoot from project namespace
+		// TODO: a project's spec.namespace can be nil, what to do about those cases?
```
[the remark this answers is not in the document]
xrstf: Should we reject such projects from targeting, telling the user to please wait until the project is completely created?

pkg/target/manager.go
xrstf: Technically yes, however I do not want the caller to figure out what namespace to use. Everything in the manager talks about projects and seeds and gardens, so I wanted to keep the naming scheme consistent. For example giving "the project _namespace_ but the seed _object_" just feels weird to me.

pkg/target/manager.go
xrstf: Sure, I just replicated the v1 behaviour. Not sure if we want to deviate already from the old behaviour.

pkg/target/manager.go
xrstf: TODO: how?

pkg/target/manager.go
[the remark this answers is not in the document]
petersutter: yes, makes sense

pkg/cmd/target/options.go
xrstf: Let's split this after this PR has been merged. The split should not affect the general structure of the `target` package and with this PR merged, we have a basis for other commands to be implemented.

pkg/cmd/root/command.go
```diff
@@ -0,0 +1,132 @@
...
+		// Use config file from the flag.
+		viper.SetConfigFile(factory.ConfigFile)
+	} else {
+		// Find home directory.
+		home, err := homedir.Dir()
+		cobra.CheckErr(err)
+
+		configPath := filepath.Join(home, gardenHomeFolder)
+
+		// Search config in $HOME/.garden or in path provided with the env variable GL_HOME with name ".garden-login" (without extension) or name from env variable GL_CONFIG_NAME.
```
petersutter, line 86: there are some more places with `GL_ prefix` that needs to be replaced with `GCTL´

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+	project := &gardencorev1beta1.Project{}
+	key := types.NamespacedName{Name: projectName}
+	err := gardenClient.Get(ctx, key, project)
+
+	return project, err
+}
+
+func (m *managerImpl) validateProject(ctx context.Context, project *gardencorev1beta1.Project) error {
+	if project.Spec.Namespace == nil || *project.Spec.Namespace == "" {
+		return errors.New("project has not yet been fully created")
```
petersutter, line 124:
```suggestion
		return errors.New("project does not have a corresponding namespace set. Most likely it has not yet been fully created")
```
so that the users knows what went wrong, in case they want to debug

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+		t.Seed = ""
+		t.Project = projectName
+		t.Shoot = ""
+
+		return nil
+	})
+}
+
+func (m *managerImpl) resolveProjectName(ctx context.Context, gardenClient client.Client, projectName string) (*gardencorev1beta1.Project, error) {
+	// validate that the project exists
```
petersutter, line 114:
```suggestion
```

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+
+		t.Seed = ""
+		t.Project = projectName
+		t.Shoot = ""
+
+		return nil
+	})
+}
+
+func (m *managerImpl) resolveProjectName(ctx context.Context, gardenClient client.Client, projectName string) (*gardencorev1beta1.Project, error) {
```
petersutter, line 113: shouldn't such a function be named `resolveProjectByName`?

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+
+	// filter found shoots
+	matchingShoots := []*gardencorev1beta1.Shoot{}
+	for i, s := range shootList.Items {
+		if s.Name != shootName {
+			continue
+		}
+
+		// if filtering by seed, ignore shoot's whose seed name doesn't matchh
```
petersutter, line 307:
```suggestion
		// if filtering by seed, ignore shoot's whose seed name doesn't match
```

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+		if err := gardenClient.Get(ctx, key, shoot); err != nil {
+			return nil, nil, nil, fmt.Errorf("invalid shoot %q: %w", key.Name, err)
+		}
+
+		return project, nil, shoot, nil
+	}
+
+	// list all shoots, filter by their name and possibly spec.seedName (if seed is set)
+	shootList := gardencorev1beta1.ShootList{}
+	if err := gardenClient.List(ctx, &shootList, &client.ListOptions{}); err != nil {
```
petersutter, line 296: use `spec.seedName` field-selector, in case we have a seed name

pkg/target/manager.go
```diff
@@ -0,0 +1,527 @@
...
+		if s.Name != shootName {
+			continue
+		}
+
+		// if filtering by seed, ignore shoot's whose seed name doesn't matchh
+		if seed != nil && (s.Status.SeedName == nil || *s.Status.SeedName != seed.Name) {
+			continue
+		}
+
+		matchingShoots = append(matchingShoots, &shootList.Items[i])
```
petersutter, line 312: why 
`matchingShoots = append(matchingShoots, &shootList.Items[i])`
and not 
`matchingShoots = append(matchingShoots, &s)`
?

pkg/target/manager.go
xrstf: Thhe function resolves the name to a project. "resolve project name" makes total sense to me. With a "by" I would expect a function name like "_find_ project _by_ name".

pkg/target/manager.go
xrstf: Do we filter by `spec.seedName` or `status.seedName`? Currently I'm filtering by status, but maybe that was wrong?

pkg/target/manager.go
xrstf: Because `s` is going to be reused in each loop interation. See https://play.golang.org/p/1YgZAv5JWNN -- using the loop variable's address is a common Go error that bit me many, many times.

petersutter: oops, bit me too! :D

pkg/target/config.go
petersutter: as discussed we will takle this later..

pkg/target/manager.go
```diff
@@ -0,0 +1,534 @@
...
+
+	// given how fast we can resolve shoots by project and that shoots
+	// always have a project, but not always a seed (yet), we prefer
+	// for users later to use the project path in their target
+	projectList := &gardencorev1beta1.ProjectList{}
+	if err := gardenClient.List(ctx, projectList, client.MatchingFields{gardencore.ProjectNamespace: shoot.Namespace}); err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to fetch parent project for shoot: %v", err)
+	}
+
+	// see note about on why we have to filter again because ctrl-runtime doesn't support FieldSelectors in tests
```
petersutter, line 350:
```suggestion
	// see note above on why we have to filter again because ctrl-runtime doesn't support FieldSelectors in tests
```

petersutter:
@xrstf there are some lint errors that need to be fixed, otherwise lgtm

petersutter:
I just tested it locally and ran into an error because the home folder was not resolved for the garden kubeconfig.
```yaml
cat ~/.garden/gardenctl-v2.yaml
gardens:
- name: my-garden
  kubeconfig: ~/path/to/kubeconfig.yaml 
```

xrstf:
@petersutter linted and handled ~ in configuration now.

pkg/cmd/root/command.go
```diff
@@ -0,0 +1,132 @@
...
+	home := os.Getenv(envGardenHomeDir)
+	if len(home) == 0 {
+		home, err = homedir.Dir()
+		cobra.CheckErr(err)
+
+		home = filepath.Join(home, gardenHomeFolder)
+	}
+
+	// prefer -session, then an explicit GCTL_HOME env,
+	// but fallback to the system-defined home directory
```
petersutter, line 120: > prefer -session

do not understand

xrstf: Yeah I've removed the notion of "sessions" again now. The comment was .. inspired, yet misleading. I am not actually sure how we want to allow users to switch between targets (we probably want to offer that, but _how_ exactly?), so it's probably best if we do it in another PR.

pkg/target/manager.go
petersutter: as discussed, let's remove it for now

petersutter (approved):
/lgtm, let's get this basic target handling in and improve as we go

PR #20, opened 2021-06-23
Return to ctrl-runtime 0.8 to be Gardener 1.25 compatible

xrstf:
I previously went with ctrl-runtime 0.9, because we needed more packages from k8s.io/cli-runtime, which in turn needed to be 0.21.x because otherwise the kustomize dependency was broken.

We now decided that this is simply not worth it, as it prevented us from upgrading Gardener and use stuff introduced with the GEP-15 PRs. With the new threadsafe buffers required to make the unit tests work, it is now acceptable to copy the entire IOStreams struct over.

This PR does exactly that, then reverts the "too brave" dependencies and then updates the Gardener dependency, so we can make use of the first chunk of newly introduced constants.

**Release note**:

```feature user
NONE
```


petersutter (approved):
/lgtm

PR #21, opened 2021-06-24
Add auto-completion for global targeting flags

### LINKED ISSUE #22: Auto completion for targeting flags
Auto completion should be added for the targeting flags `--garden`, `--project`, `--seed`, `--shoot` 

**Why is this needed**:
To support the user when targeting by providing completion suggestions.

### LINKED ISSUE #23: Auto completion for target arguments
Auto completion should be added for the target arguments `garden`, `project`, `seed`, `shoot`, similar to what needs to be done with #22 
e.g. `gardenctl target --project myproject shoot <tab>`
this should list the shoot names in project myproject of the targeted `garden` cluster

**Why is this needed**:
To support the user when targeting by providing completion suggestions.

xrstf:
This PR adds auto-completion to `--garden`, `--project`, `--seed` and `--shoot`. It also slightly changes the kubeconfig caching behaviour, which previously only fetched the kubeconfig when doing a `gardenctl target ...`, but now it fetches the kubeconfig whenever it's needed (so that `gardenctl --garden dev --shoot foo ssh` can work without targeting the shoot "foo" first).

This PR also adds the related auto-completion for the `target` command.

So far, errors that happen during the completion are not shown anywhere. I guess stderr is swallowed somewhere somehow?

Fixes #22, fixes #23

**Release note**:

```feature user
Add auto-completion to `--garden`, `--project`, `--seed` and `--shoot`
```


pkg/cmd/root/completion.go
```diff
@@ -0,0 +1,268 @@
...
+	} else if targetProvider.SeedNameFlag != "" {
+		currentTarget = currentTarget.WithSeedName(targetProvider.SeedNameFlag).WithProjectName("")
+	}
+
+	gardenClient, err := manager.GardenClient(currentTarget)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create Kubernetes client for garden cluster %q: %w", currentTarget.GardenName(), err)
+	}
+
+	ctx := context.Background()
```
petersutter, line 236: Use context from factory?

pkg/cmd/root/completion.go
```diff
@@ -0,0 +1,268 @@
...
+		factory.TargetFile = targetProvider.TargetFile
+		factory.TargetProvider = nil
+
+		result, err := completer(cmd, args, toComplete)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%v\n", err)
+			return nil, cobra.ShellCompDirectiveDefault
+		}
+
+		return result, cobra.ShellCompDirectiveDefault
```
petersutter, line 112: use `ShellCompDirectiveNoFileComp` instead (or in addition?)

pkg/cmd/target/completion.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	manager, err := f.Manager()
+	if err != nil {
+		return nil, err
+	}
+
+	// NB: this uses the DynamicTargetProvider from the root cmd and
+	// is therefore aware of flags like --garden; the goal here is to
+	// allow the user to type "gardenctl target --garden [tab][select] --project [tab][select] shoot [tab][select]"
```
petersutter, line 43: allow to target also only with flags
`gardenctl target --garden [tab][select] --project [tab][select] --shoot [tab][select]`
`Error: failed to complete command options: expected exactly 2 arguments`

internal/util/target.go
```diff
@@ -90,6 +91,36 @@ func shootForTargetViaSeed(ctx context.Context, gardenClient client.Client, t ta
...
+		if project.Spec.Namespace == nil {
+			return nil, nil
+		}
+
+		listOpt = &client.ListOptions{Namespace: *project.Spec.Namespace}
+	} else if t.SeedName() != "" {
+		listOpt = client.MatchingFields{gardencore.ShootSeedName: t.SeedName()}
+	} else {
+		return nil, errors.New("invalid target, must have either project or seed specified for targeting a shoot")
+	}
```
petersutter, line 114: do we really want to forbid this?

pkg/cmd/root/completion.go
```diff
@@ -0,0 +1,236 @@
...
+	if err := gardenClient.List(factory.Context(), projectList); err != nil {
+		return nil, fmt.Errorf("failed to list projects on garden cluster %q: %w", currentTarget.GardenName(), err)
+	}
+
+	names := sets.NewString()
+	for _, project := range projectList.Items {
+		names.Insert(project.Name)
+	}
+
+	return names.List(), nil
```
petersutter, line 160: once the current target is determined, this code looks the same as for the argument completion. Same for the other completions. Doesn't it make sense to de-duplicate even more code?

pkg/cmd/root/completion.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func shootFlagCompletionFunc(cmd *cobra.Command, args []string, toComplete string) ([]string, error) {
+	manager, err := factory.Manager()
+	if err != nil {
+		return nil, fmt.Errorf("failed to create manager: %w", err)
+	}
+
+	// for simplicity, always try to read the current target file and ignore any errors
```
petersutter, line 205: This comment makes sense when you know how it was previously done. If you do not know the context, you would think that you just spared the error handling as you were too lazy (which was not the case of course) ;-)

petersutter:
Please also add tests for the completions

internal/util/target.go
xrstf: How do you mean? The "you must target a project or seed when you want to target a shoot"-logic is ingrained everywhere in the targeting.

pkg/cmd/root/completion.go
xrstf: I don't quite understand what you code blocks you mean.

petersutter: https://github.com/gardener/gardenctl-v2/pull/21/files/240f1aeb8acd2c5ee9f78cb9966692d53940b9c0#diff-517d5acc56e30d22994a8901475305ae60213bb2598de53d75c92da4e14d4bc8R145-R160
```go
	gardenClient, err := manager.GardenClient(currentTarget)
	if err != nil {
		return nil, fmt.Errorf("failed to create Kubernetes client for garden cluster %q: %w", currentTarget.GardenName(), err)
	}

	projectList := &gardencorev1beta1.ProjectList{}
	if err := gardenClient.List(factory.Context(), projectList); err != nil {
		return nil, fmt.Errorf("failed to list projects on garden cluster %q: %w", currentTarget.GardenName(), err)
	}

	names := sets.NewString()
	for _, project := range projectList.Items {
		names.Insert(project.Name)
	}

	return names.List(), nil
```

https://github.com/gardener/gardenctl-v2/pull/21/files/240f1aeb8acd2c5ee9f78cb9966692d53940b9c0#diff-ace14d9438fa14f933287b67235812a949d856032fd22ea809868e74679f1ccdR76-R91
```go
	gardenClient, err := manager.GardenClient(t)
	if err != nil {
		return nil, fmt.Errorf("failed to create Kubernetes client for garden cluster %q: %w", t.GardenName(), err)
	}

	projectList := &gardencorev1beta1.ProjectList{}
	if err := gardenClient.List(ctx, projectList); err != nil {
		return nil, fmt.Errorf("failed to list projects on garden cluster %q: %w", t.GardenName(), err)
	}

	names := sets.NewString()
	for _, project := range projectList.Items {
		names.Insert(project.Name)
	}

	return names, nil
```

internal/util/target.go
petersutter: Already now I can target a shoot without targeting a project or seed previously
``` bash
$ gardenctl target garden landscape-dev
Successfully targeted garden "landscape-dev".
```
```bash
$ gardenctl target shoot myshoot`
`Successfully targeted shoot "myshoot".
```
and it will automagically figure out the project
```bash
cat ~/.garden/target.yaml
garden: landscape-dev
project: myproject
shoot: myshoot
```

xrstf: Okay then, I will refactor everything.

xrstf: Actually, we should do this in a separate PR. Changing this core aspect will make it really hard to review otherwise.

petersutter: ok fine with me

xrstf:
> However running target command without any flags returns this which is somewhat unexpected

That's an ugly sid effect of the way we transparently handle the CLI flags for tageting. The `target` command could not distinguish between explicit information (CLI flags/args) and implicit information (`target.yaml`).

I tried a few different approaches to make this distinction available to the target command, but none of them turned out too great. Eventually I decided on simply making it possible to disable loading the `target.yaml`. This has the advantage of not totally breaking the "want the target? ask the `TargetProvider`, don't look at the flags yourself!"-contract we have (in the form of the `TargetProvider` interface), while still cheating and telling the `target` command explicitly that the `TargetProvider` is not generic, but indeed a `DynamicTargetProvider`.

The impact of this tiny cheat is IMHO totally worth it. The `target` command should be the only special one in this regard.

petersutter (approved):
/lgtm
let's tackle the open point in a different pr, as discussed here https://github.com/gardener/gardenctl-v2/pull/21#discussion_r659833429

PR #42, opened 2021-10-07
Bug/targeting changes

grolu:
With this PR we change how the targeting behaves if a seed is targeted.
If a `seed` is targeted and the user targets a `shoot`, the seed is discarded in favor for `project` and `shoot`.
Also, if the user targets a `seed`, `project` and `shoot` are discarded.
This ensures that `shoot` and `seed` are never targeted at the same time.

This PR also renames the `target drop` command to `target unset`

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/pkg/target/manager.go
+++ b/pkg/target/manager.go
@@ -35,7 +35,7 @@ type Manager interface {
 	TargetFlags() TargetFlags
 
 	// TargetGarden sets the garden target configuration
-	// This implicitly drops project, seed and shoot target configuration
+	// This implicitly unsets project, seed and shoot target configuration
 	TargetGarden(name string) error
 	// TargetProject sets the project target configuration
 	// This implicitly unsets seed and shoot target configuration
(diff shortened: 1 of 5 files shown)


pkg/target/manager.go
```diff
@@ -307,11 +304,11 @@ func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
 		} else if t.Seed != "" {
 			seed, err = m.resolveSeedName(ctx, gardenClient, t.Seed)
 			if err != nil {
-				return fmt.Errorf("failed to fetch kubeconfig for seed cluster: %w", err)
+				return fmt.Errorf("failed to resolve sed by name: %w", err)
```
petersutter, line 307:
```suggestion
				return fmt.Errorf("failed to resolve seed by name: %w", err)
```

petersutter: however this error message is not consistent with the error message for `resolveShootName`, which is `failed to resolve shoot`

pkg/target/manager_test.go
```diff
@@ -461,36 +446,36 @@ var _ = Describe("Manager", func() {
...
 		Expect(err).NotTo(HaveOccurred())
 		Expect(manager).NotTo(BeNil())
 
-		// Drop Project
-		Expect(manager.DropTargetProject()).Should(Equal(prod1Project.Name))
+		// Unset Project
+		Expect(manager.UnsetTargetProject()).Should(Equal(prod1Project.Name))
 
-		// should also drop seed and shoot
+		// should also unset seed and shoot
```
petersutter, line 466: in this test, seed was never targeted
```suggestion
		// should also unset shoot
```

PR #43, opened 2021-10-08
Enh/domain targeting

### LINKED ISSUE #38: Implement Domain Targeting
### Domain Targetting
Domain targetting extracts the domain from an API server or Gardener Dashboard URL and matches it against the domain secrets in the garden namespaces of all configured gardens (pre-fetched, of course) or accesses the cluster-identity configmap in the kube-system namespace of the shoot if the shoot uses a custom, i.e. unknown domain (slower, therefore only second option).

```
kubectl [seed|shoot] https://api.funny.core.shoot.prod.com                                  # target seed or shoot by API server URL
kubectl [seed|shoot] https://dashboard.garden.prod.com/namespace/garden-core/shoots/funny   # target seed or shoot by Gardener Dashboard URL
```

grolu:
This PR introduces basic domain targeting.
Shoots can be targeted via dashboard url or shoot api server url.
To enable domain targeting, you have to extend your garden configuration:

```
gardens:
  - name: foo-garden
    kubeconfig: ~/.kube/foo-kubeconfig
    shootDomain: shoot.gardener.cloud #shoot API sever domain
    dashboardDomain: dashboard.gardener.cloud #url to gardener dashboard
```
Fixes #38

**Release note**:

```feature user

```

### DIFF
--- a/pkg/target/manager.go
+++ b/pkg/target/manager.go
@@ -332,6 +332,10 @@ func targetFlagsFromDomain(ctx context.Context, m *managerImpl, t *targetImpl, n
 		return nil, fmt.Errorf("invalid domain format in configured gardens: %w", err)
 	}
 
+	if len(dashboardDomains) == 0 && len(shootDomains) == 0 {
+		return nil, fmt.Errorf("no domain found in configured gardens. Shoot targeting via domain requires shoot and / or dashboard domain configuration for each garden that you want to enable domain targeting for")
+	}
+
 	garden, domain := matchGardenDomain(dashboardDomains, name)
 	if domain != "" {
 		domainRegexp, err := regexp.Compile(`.+/namespace/([^/]+)/shoots/([^/]+)/?`)
@@ -401,7 +405,7 @@ func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
 				return nil
 			}
 
-			return fmt.Errorf("failed to target shoot with url, %s", shootName)
+			return fmt.Errorf("failed to target shoot with url: %s", shootName)
 		}
 
 		if t.Garden == "" {


pkg/config/config.go
```diff
@@ -19,8 +19,10 @@ type Config struct {
 }
 
 type Garden struct {
-	Name       string `yaml:"name"`
-	Kubeconfig string `yaml:"kubeconfig"`
+	Name            string `yaml:"name"`
+	Kubeconfig      string `yaml:"kubeconfig"`
+	ShootDomain     string `yaml:"shootDomain"`
+	DashboardDomain string `yaml:"dashboardDomain"`
```
holgerkoser, line 25: Why don't we add a generic Patter Matching mechanism which also can handle technicalID. Add an Array of e.g `shootRegexps` or `shootMatchPatterns`. The implementation should be generic and could look like https://play.golang.org/p/q7JwrHDu283.

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+	for garden, gDomain := range domains {
+		if strings.Contains(domain, gDomain) {
+			return garden, domain
+		}
+	}
+
+	return "", ""
+}
+
+func targetFlagsFromDomain(ctx context.Context, m *managerImpl, t *targetImpl, name string) (TargetFlags, error) {
```
holgerkoser, line 329: This should be a method of the managerImp. It should return a Target and not a TargetFlags. Why does it have in input parameter of type targteImpl. There is no need to depend on an impl of an interface. The only info we use is the gardenName.

pkg/config/config.go
grolu: Yes, this sounds like a good Idea. We should document this well, so that users know how to configure their gardens. Would you add this per garden or global? If we have separate regexps per garden, we could keep the functionality of also targeting across multiple gardens with shoot / dashboard urls. Maybe we can have both?

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+		match := domainRegexp.FindStringSubmatch(domain)
+		if match != nil {
+			return NewTargetFlags(garden, match[2], "", match[1]), nil
+		}
+	}
+
+	return nil, nil
+}
+
+func isDomainFormat(name string) bool {
```
holgerkoser, line 387: Why not `strings.Contains`

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
 	return nil
 }
 
+func domainsFromConfiguration(m *managerImpl) (map[string]string, map[string]string, error) {
```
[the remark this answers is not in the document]
grolu: probably this happened during refactoring. I will fix this.

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+}
+
+func targetFlagsFromDomain(ctx context.Context, m *managerImpl, t *targetImpl, name string) (TargetFlags, error) {
+	dashboardDomains, shootDomains, err := domainsFromConfiguration(m)
+	if err != nil {
+		return nil, fmt.Errorf("invalid domain format in configured gardens: %w", err)
+	}
+
+	if len(dashboardDomains) == 0 && len(shootDomains) == 0 {
+		return nil, fmt.Errorf("no domain found in configured gardens. Shoot targeting via domain requires shoot and / or dashboard domain configuration for each garden that you want to enable domain targeting for")
```
petersutter, line 336: if you need to format your error message use `fmt.Errorf()`, if you don't use `errors.New()`.
-> in this case, use `errors.New()`

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+			}
+
+			namespace, err := m.resolveProjectNamespace(ctx, gardenClient, match[1])
+			if err != nil {
+				return nil, fmt.Errorf("failed to validate project: %w", err)
+			}
+
+			if namespace == nil {
+				return nil, fmt.Errorf("invalid namespace in shoot url")
+			}
```
petersutter, line 360: I don't think this can ever happen

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+	garden, domain = matchGardenDomain(shootDomains, name)
+	if domain != "" {
+		domainRegexp, err := regexp.Compile(`([^.]+).([^.]+)\..+`)
+		if err != nil {
+			return nil, fmt.Errorf("could not extract project and shoot from url: %w", err)
+		}
+
+		match := domainRegexp.FindStringSubmatch(domain)
+		if match != nil {
+			return NewTargetFlags(garden, match[2], "", match[1]), nil
```
petersutter, line 380: I think it makes it hard to read what match[2] and match[1] really means, it probably makes more sense to extract it to a variable and name it appropriately

pkg/target/manager.go
```diff
@@ -282,8 +292,122 @@ func (m *managerImpl) validateSeed(ctx context.Context, seed *gardencorev1beta1.
...
+			}
+
+			if targetFlags != nil {
+				t.Garden = targetFlags.GardenName()
+				t.Project = targetFlags.ProjectName()
+				t.Shoot = targetFlags.ShootName()
+				return nil
+			}
+
+			return fmt.Errorf("failed to target shoot with url: %s", shootName)
```
petersutter, line 408: Nit: I'd prefer to turn it around

```suggestion
			if targetFlags == nil {
				return fmt.Errorf("failed to target shoot with url: %s", shootName)
			}

			t.Garden = targetFlags.GardenName()
			t.Project = targetFlags.ProjectName()
			t.Shoot = targetFlags.ShootName()
			
			return nil
```

PR #60, opened 2021-12-23
Implemented shoot control plane targeting

grolu:
This PR allows to set a `--controlplane` flag while targeting.
This flag allows to set a controlplane flag on the current target. This only works if a shoot is already targeted or provided along with the flag.
If set, this flag will influence the kubeconfig command to provide a kubeconfig for the shoot's control plane (seed with appropriate namespace selected).
Currently, there is no way to explicitly unset the flag. It is possible to re-target the shoot without the `--controlplane` flag though. Moreover, the flag will be removed implicitly when the targeted shoot gets dropped, by selecting a different garden / project or by unsetting the shoot with the `unset` command.

Fixes #

We need to discuss if we want to introduce a dedicated `controlplane` subcommand. Right now, the functionality is kind of hidden in the (global) target flags. Also, a dedicated command would allow to set / unset the flag directly.

NOTE: There is currently no real test for this functionality other than the basic target flag tests as we as of no have no tests in place that test the targeting command with the target flags. I did not invest here as we need to discuss parts of the feature anyway.

**Release note**:

```feature user

```


pkg/target/target.go
```diff
@@ -57,22 +62,24 @@ type Target interface {
...
+	Shoot        string `yaml:"shoot,omitempty" json:"shoot,omitempty"`
+	ControlPlane bool   `yaml:"controlPlane,omitempty" json:"controlPlane,omitempty"`
 }
 
 var _ Target = &targetImpl{}
 
 // NewTarget returns a new target. This function does not perform any validation,
 // so the returned target can be invalid.
-func NewTarget(gardenName, projectName, seedName, shootName string) Target {
+func NewTarget(gardenName, projectName, seedName, shootName string, shootControlPlane bool) Target {
```
grolu, line 76: Note (just for me to remember if this question comes up) - I wanted to add a check to the `Validate()` method, something like
```
if t.ControlPlane && len(t.Shoot) == 0 {
    return errors.New("control plane can only be configured if a shoot is targeted")
}
```
However, this was not possible as this would prevent the user from targeting a shoot control plane like this:
`gardenctl target shoot foo --controlplane`
As the shoot is not set on the `currenttarget` at the point of `Validate()` function invocation.

There will be an error thrown by the `targetBuilder` anyway when trying to target a control plane without targeting a shoot before or along with the command.

holgerkoser, line 76: Do not change the signature of the `NewTarget` function which would result in the `ControlPlane` property to be `false`. Add a new function called `NewControlPlaneTarget` with the same parameters but `ControlPlane` property set to true or use the WithControlPlane method where needed.

pkg/target/target.go
```diff
@@ -35,6 +35,8 @@ type Target interface {
 	SeedName() string
 	// ShootName returns the currently targeted shoot cluster name.
 	ShootName() string
+	// ShootControlPlane returns true if shot control plane is targeted.
+	ShootControlPlane() bool
```
holgerkoser, line 39: Why not `ControlPlane`?

pkg/cmd/target/target.go
```diff
@@ -240,7 +251,19 @@ func (o *TargetOptions) Complete(f util.Factory, cmd *cobra.Command, args []stri
 }
 
 // Validate validates the provided options
-func (o *TargetOptions) Validate() error {
+func (o *TargetOptions) Validate(f util.Factory) error {
```
holgerkoser, line 254: You broke the interface contract with this change. Validate does not have args!!! Everthing you need for validation needs to be added to the options in complete. That's the idea.

pkg/target/manager.go
```diff
@@ -319,6 +342,10 @@ func (m *managerImpl) TargetMatchPattern(ctx context.Context, value string) erro
 		tb.SetShoot(ctx, tm.Shoot)
 	}
 
+	if m.TargetFlags().ControlPlane() {
+		tb.SetShootControlPlane(ctx, m.TargetFlags().ControlPlane())
```
holgerkoser, line 346:
```suggestion
		tb.SetControlPlane(ctx, m.TargetFlags().ControlPlane())
```

pkg/target/manager.go
```diff
@@ -254,7 +258,25 @@ func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
...
+	tb := NewTargetBuilder(m.config, m.clientProvider)
+
+	currentTarget, err := m.CurrentTarget()
+	if err != nil {
+		return fmt.Errorf("failed to get current target: %v", err)
+	}
+
+	tb.Init(currentTarget)
+
+	target, err := tb.SetShootControlPlane(ctx, m.TargetFlags().ControlPlane()).Build()
```
holgerkoser, line 279:
```suggestion
	target, err := tb.SetControlPlane(ctx, m.TargetFlags().ControlPlane()).Build()
```

pkg/target/manager.go
```diff
@@ -254,7 +258,25 @@ func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
 
 	tb.Init(currentTarget)
 
-	target, err := tb.SetShoot(ctx, shootName).Build()
+	target, err := tb.SetShoot(ctx, shootName).SetShootControlPlane(ctx, m.TargetFlags().ControlPlane()).Build()
```
holgerkoser, line 261:
```suggestion
	target, err := tb.SetShoot(ctx, shootName).SetControlPlane(ctx, m.TargetFlags().ControlPlane()).Build()
```

pkg/target/manager.go
```diff
@@ -333,6 +360,7 @@ func (m *managerImpl) patchTargetWithTarget(target Target) error {
 		t.Project = target.ProjectName()
 		t.Seed = target.SeedName()
 		t.Shoot = target.ShootName()
+		t.ControlPlane = target.ShootControlPlane()
```
holgerkoser, line 363:
```suggestion
		t.ControlPlane = target.ControlPlane()
```

pkg/target/target.go
```diff
@@ -106,28 +113,39 @@ func (t *targetImpl) ShootName() string {
 	return t.Shoot
 }
 
+// ShootControlPlane returns true if shot control plane is targeted.
+func (t *targetImpl) ShootControlPlane() bool {
```
holgerkoser, line 117:
```suggestion
func (t *targetImpl) ControlPlane() bool {
```

pkg/target/target.go
```diff
@@ -106,28 +113,39 @@ func (t *targetImpl) ShootName() string {
...
+	return t.ControlPlane
+}
+
 // WithGardenName returns a copy of the target with the garden name updated.
 // The returned target can be invalid.
 func (t *targetImpl) WithGardenName(name string) Target {
-	return NewTarget(name, t.Project, t.Seed, t.Shoot)
+	return NewTarget(name, t.Project, t.Seed, t.Shoot, t.ControlPlane)
```
holgerkoser, line 124:
```suggestion
	return newTargetImpl(name, t.Project, t.Seed, t.Shoot, t.ControlPlane)
```

pkg/target/target_builder.go
```diff
@@ -33,6 +33,8 @@ type TargetBuilder interface {
 	SetSeed(context.Context, string) TargetBuilder
 	// SetShoot updates TargetBuilder with a Shoot name
 	SetShoot(context.Context, string) TargetBuilder
+	// SetShootControlPlane updates TargetBuilder shoot control plane flag
+	SetShootControlPlane(context.Context, bool) TargetBuilder
```
holgerkoser, line 37:
```suggestion
	SetControlPlane(context.Context, bool) TargetBuilder
```

pkg/target/target_flags.go
```diff
@@ -37,20 +39,22 @@ type TargetFlags interface {
...
 type targetFlagsImpl struct {
-	gardenName  string
-	projectName string
-	seedName    string
-	shootName   string
+	gardenName   string
+	projectName  string
+	seedName     string
+	shootName    string
+	controlPlane *bool
```
holgerkoser, line 57: Should be bool not pointer to bool

holgerkoser, line 57:
```suggestion
	controlPlane bool
```

pkg/target/target_flags.go
```diff
@@ -69,19 +73,24 @@ func (tf *targetFlagsImpl) ShootName() string {
...
+func (tf *targetFlagsImpl) ControlPlane() bool {
+	return *tf.controlPlane
+}
+
 func (tf *targetFlagsImpl) AddFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&tf.gardenName, "garden", "", "target the given garden cluster")
 	flags.StringVar(&tf.projectName, "project", "", "target the given project")
 	flags.StringVar(&tf.seedName, "seed", "", "target the given seed cluster")
 	flags.StringVar(&tf.shootName, "shoot", "", "target the given shoot cluster")
+	flags.BoolVar(tf.controlPlane, "controlplane", *tf.controlPlane, "target control plane of shoot, use together with shoot argument")
```
holgerkoser, line 85:
```suggestion
	flags.BoolVar(&tf.controlPlane, "controlplane", tf.controlPlane, "target control plane of shoot, use together with shoot argument")
```

pkg/target/target_flags.go
```diff
@@ -69,19 +73,24 @@ func (tf *targetFlagsImpl) ShootName() string {
...
 }
 
 func (tf *targetFlagsImpl) ToTarget() Target {
-	return NewTarget(tf.gardenName, tf.projectName, tf.seedName, tf.shootName)
+	return NewTarget(tf.gardenName, tf.projectName, tf.seedName, tf.shootName, *tf.controlPlane)
```
holgerkoser, line 89:
```suggestion
	return NewTarget(tf.gardenName, tf.projectName, tf.seedName, tf.shootName, tf.controlPlane)
```

pkg/target/target_flags.go
```diff
@@ -69,19 +73,24 @@ func (tf *targetFlagsImpl) ShootName() string {
...
 }
 
 func (tf *targetFlagsImpl) isEmpty() bool {
-	return tf.gardenName == "" && tf.projectName == "" && tf.seedName == "" && tf.shootName == ""
+	return tf.gardenName == "" && tf.projectName == "" && tf.seedName == "" && tf.shootName == "" && !*tf.controlPlane
```
holgerkoser, line 93: We should consider it to be empty even if the controlPlane is true

pkg/target/target_flags.go
```diff
@@ -112,6 +121,8 @@ func (tf *targetFlagsImpl) OverrideTarget(current Target) (Target, error) {
 			current = current.WithShootName(tf.shootName)
 		}
 
+		current = current.WithControlPlane(*tf.controlPlane)
```
holgerkoser, line 124:
```suggestion
		current = current.WithControlPlane(tf.controlPlane)
```

pkg/target/target_flags.go
```diff
@@ -37,20 +39,22 @@ type TargetFlags interface {
...
 	return &targetFlagsImpl{
-		gardenName:  garden,
-		projectName: project,
-		seedName:    seed,
-		shootName:   shoot,
+		gardenName:   garden,
+		projectName:  project,
+		seedName:     seed,
+		shootName:    shoot,
+		controlPlane: &controlPlane,
```
holgerkoser, line 48:
```suggestion
		controlPlane: controlPlane,
```

pkg/target/target_flags.go
petersutter: I was wondering the same

petersutter: .. same for https://github.com/gardener/gardenctl-v2/blob/3b506a41fc8e4459764f82026717e3f77ea98a8b/pkg/target/target.go#L183

pkg/target/manager.go
grolu: Yes, I also struggled with this naming. At some parts I added _Shoot_ to the function names to make it clear that we target Shoot Control Plane, not e.g. the one of the Seed. But for now and to align things I'm also happy to remove it, which I'll do.

pkg/target/target.go
grolu: see comment above. But as I already wrote, and now with some distance to my work, I agree that we should remove the _Shoot_ term. The flag is also not called `--shootcontrolplane` so let's keep it aligned.

grolu: But we need to add some kind of suffix as we did for the other functions where we added `Name` as it would otherwise clash with the struct names. I decided to call It `ControlPlaneFlag`

pkg/target/target.go
grolu: ok

pkg/target/target_flags.go
grolu: done

pkg/target/target_flags.go
grolu: I changed this back and forth several times during development. There were some cases where the targeting did not work, e.g. `target --controlplane` without specifying other target params, however I was not able to reproduce it. Therefore I removed `controlplane` checks from the isEmpty functions. If we encounter issues we can re-add it.

pkg/cmd/target/target.go
```diff
@@ -85,7 +87,11 @@ func runCmdTarget(f util.Factory, o *TargetOptions) error {
 	}
 
 	if o.Output == "" {
-		fmt.Fprintf(o.IOStreams.Out, "Successfully targeted %s %q\n", o.Kind, o.TargetName)
+		if o.Kind == TargetKindControlPlane {
```
holgerkoser, line 90: The currentTarget error above should be wrapped with %w

pkg/target/manager.go
```diff
@@ -63,6 +66,8 @@ type Manager interface {
 	UnsetTargetSeed() (string, error)
 	// UnsetTargetShoot unsets the garden shoot configuration
 	UnsetTargetShoot() (string, error)
+	// UnsetTargetControlPlane unsets the control plane flag
+	UnsetTargetControlPlane() (string, error)
```
holgerkoser, line 70: I don't understand why this method returns the targeted shoot. All other methods return the previous value which has been unset on the target object. I think this method should only return `errror` or `(bool, error)` but than bool value would always be true if no error occurs.

pkg/target/manager.go
```diff
@@ -291,6 +299,42 @@ func (m *managerImpl) UnsetTargetShoot() (string, error) {
...
+	if currentTarget.ControlPlaneFlag() {
+		return targetedName, m.patchTarget(func(t *targetImpl) error {
+			t.ControlPlane = false
+
+			return nil
+		})
+	}
+
+	return "", ErrNoControlPlaneTargeted
+}
```
holgerkoser, line 336:
```suggestion
func (m *managerImpl) UnsetTargetControlPlane() error {
	currentTarget, err := m.CurrentTarget()
	if err != nil {
		return fmt.Errorf("failed to get current target: %v", err)
	}

	if currentTarget.ControlPlaneFlag() {
		return m.patchTarget(func(t *targetImpl) error {
			t.ControlPlane = false

			return nil
		})
	}

	return ErrNoControlPlaneTargeted
}
```

pkg/target/target.go
```diff
@@ -35,6 +35,8 @@ type Target interface {
 	SeedName() string
 	// ShootName returns the currently targeted shoot cluster name.
 	ShootName() string
+	// ControlPlaneFlag returns true if shoot control plane is targeted.
+	ControlPlaneFlag() bool
```
holgerkoser, line 39:
```suggestion
	// ControlPlane returns true if shoot control plane is targeted.
	ControlPlane() bool
```

pkg/target/target_flags.go
```diff
@@ -115,6 +124,8 @@ func (tf *targetFlagsImpl) OverrideTarget(current Target) (Target, error) {
 		if err := current.Validate(); err != nil {
 			return nil, fmt.Errorf("invalid target flags: %w", err)
 		}
+
+		current = current.WithControlPlane(tf.controlPlane)
```
holgerkoser, line 128: Why is this not done before `current.Validate`?
Why do we always override the `control-plane` even if not specified in target flags?

pkg/target/manager.go
grolu: changed interface

pkg/target/target_flags.go
grolu: You are right, this is a bug introduced recently while I was experimenting with the target flags. Fixed it.

pkg/cmd/target/unset.go
```diff
@@ -62,7 +64,12 @@ func runCmdUnset(f util.Factory, o *UnsetOptions) error {
 	case TargetKindShoot:
 		targetName, err = manager.UnsetTargetShoot()
 	case TargetKindControlPlane:
-		targetName, err = manager.UnsetTargetControlPlane()
+		err = manager.UnsetTargetControlPlane()
+		if err == nil {
+			var currentTarget target.Target
+			currentTarget, err = manager.CurrentTarget()
+			targetName = currentTarget.ShootName()
+		}
```
petersutter, line 72:
```suggestion
		var currentTarget target.Target
		currentTarget, err = manager.CurrentTarget()
		targetName = currentTarget.ShootName()
			
		err = manager.UnsetTargetControlPlane()
```
WDYT?

pkg/target/manager.go
```diff
@@ -317,22 +317,21 @@ func (m *managerImpl) TargetControlPlane(ctx context.Context) error {
 	return m.updateTarget(target)
 }
 
-func (m *managerImpl) UnsetTargetControlPlane() (string, error) {
+func (m *managerImpl) UnsetTargetControlPlane() error {
 	currentTarget, err := m.CurrentTarget()
 	if err != nil {
-		return "", fmt.Errorf("failed to get current target: %v", err)
+		return fmt.Errorf("failed to get current target: %v", err)
```
petersutter, line 323:
```suggestion
		return fmt.Errorf("failed to get current target: %w", err)
```

pkg/target/manager.go
```diff
@@ -317,22 +317,21 @@ func (m *managerImpl) TargetControlPlane(ctx context.Context) error {
...
+	if currentTarget.ControlPlane() {
+		return m.patchTarget(func(t *targetImpl) error {
+			t.ControlPlaneFlag = false
 
 			return nil
 		})
 	}
 
-	return "", ErrNoControlPlaneTargeted
+	return ErrNoControlPlaneTargeted
```
petersutter, line 334:
```go
	if !currentTarget.ControlPlane() {
		return ErrNoControlPlaneTargeted
	}

	return m.patchTarget(func(t *targetImpl) error {
		t.ControlPlaneFlag = false

		return nil
	})
```

I would turn it around, but this is just my personal preference. Feel free to ignore

pkg/cmd/target/unset.go
holgerkoser: In both cases the err of `manager.CurrentTarget()` is not handled. 
```suggestion
		currentTarget, err := manager.CurrentTarget(); 
		if err != nil {
		  targetName = "nothing"
		   break
		}
		targetName = currentTarget.ShootName()
		err = manager.UnsetTargetControlPlane()
```

pkg/cmd/target/target.go
```diff
@@ -100,15 +106,16 @@ func runCmdTarget(f util.Factory, o *TargetOptions) error {
...
-	TargetKindGarden  TargetKind = "garden"
-	TargetKindProject TargetKind = "project"
-	TargetKindSeed    TargetKind = "seed"
-	TargetKindShoot   TargetKind = "shoot"
-	TargetKindPattern TargetKind = "pattern"
+	TargetKindGarden       TargetKind = "garden"
+	TargetKindProject      TargetKind = "project"
+	TargetKindSeed         TargetKind = "seed"
+	TargetKindShoot        TargetKind = "shoot"
+	TargetKindPattern      TargetKind = "pattern"
```
holgerkoser, line 113: We have a TargetKind "pattern" in the enum but the value is not a valid TargetKind?

pkg/cmd/target/unset.go
grolu: I don't understand ... the error is returned after the switch

pkg/target/manager.go
grolu: ok

pkg/cmd/target/target.go
grolu: What do you mean? Right now, when you do not provide garden, project, seed, shoot or control-plane. It defaults to pattern. We need this now when we introduce subcommands we can get rid of the targetkinds and pattern targeting will be used for the root command and the subcommands do the other target kinds.
Anyway this has nothing to do with this PR.

pkg/target/manager.go
```diff
@@ -53,6 +54,8 @@ type Manager interface {
 	// This implicitly unsets seed target configuration
 	// It will also configure appropriate project and seed values if not already set
 	TargetShoot(ctx context.Context, name string) error
+	// TargetControlPlane sets the control plane target flag
```
holgerkoser, line 57: I would replace all these `Target` und `UnsetTarget` methods with:
```go
    SetTarget(ctx context.Context, kind TargetKind, name string) error
    UnsetTarget(ctx context.Context, kind TargetKind) error
```
and would define the TargetKind enum in the target package:
```go
type TargetKind int

const (
	TargetKindInvalid TargetKind = iota - 1
	TargetKindAll
	TargetKindGarden
	TargetKindProject
	TargetKindSeed
	TargetKindShoot
	TargetKindControlPlane
)

var ValidTargetKinds = [...]string{"All", "Garden", "Project", "Seed", "Shoot", "ControlPlane"}

func (k TargetKind) String() string {
	if !k.IsValid() {
		return "Invalid"
	}

	return ValidTargetKinds[k]

}

func (k TargetKind) IsValid() bool {
	return k >= 0 && int(k) < len(ValidTargetKinds)
}
```

holgerkoser: In this Proposal the TargetPattern is renamed to TargetAll because UnsetPattern would not make sense in this case but UnsetAll would reset everything to an empty target

pkg/target/target.go
```diff
@@ -57,10 +62,11 @@ type Target interface {
...
 type targetImpl struct {
-	Garden  string `yaml:"garden,omitempty" json:"garden,omitempty"`
-	Project string `yaml:"project,omitempty" json:"project,omitempty"`
-	Seed    string `yaml:"seed,omitempty" json:"seed,omitempty"`
-	Shoot   string `yaml:"shoot,omitempty" json:"shoot,omitempty"`
+	Garden           string `yaml:"garden,omitempty" json:"garden,omitempty"`
+	Project          string `yaml:"project,omitempty" json:"project,omitempty"`
+	Seed             string `yaml:"seed,omitempty" json:"seed,omitempty"`
+	Shoot            string `yaml:"shoot,omitempty" json:"shoot,omitempty"`
+	ControlPlaneFlag bool   `yaml:"controlPlane,omitempty" json:"controlPlane,omitempty"`
```
holgerkoser, line 69: We could also do something like kubernetes does with TypeMeta:
```go
type shootTargetImpl {
	Name            string `json:"shoot,omitempty"`
	ControlPlane bool
}
type targetImpl struct {
	Garden           string 
	Project          string
	Seed             string
	Shoot        shootTargetImpl `json:",inline"`
}
```

pkg/cmd/target/unset.go
```diff
@@ -11,6 +11,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/gardener/gardenctl-v2/pkg/target"
+
```
holgerkoser, line 15:
```suggestion
```

pkg/cmd/target/target.go
holgerkoser: Sorry. I have missed the fact that pattern of course is in the ValidTargetKinds list. I don't knwo anymore why I thought it is invalid to call `gardenctl target pattern ...`

pkg/cmd/target/unset.go
holgerkoser:
```
currentTarget, err = manager.CurrentTarget()
targetName = currentTarget.ShootName()
```
I assume that currentTarget is `nil` if the `manager.CurrentTarget()` return an error. You cannot call `ShootName()` on a nil pointer.

holgerkoser: In golang all errors have to handled immediately. We should also never override previous errors and handle only the last one even if all the calls in between would work. If we could ignore an error we should write:
```go
currentTarget, _ = manager.CurrentTarget()
```

grolu: You are right, of course. I changed it to return the error directly. We could also just not fetch the shoot name in this case as it is only used for the output anyway. But I think if an error occurs something is wrong (file read failed) and the user should get notified.

pkg/target/manager.go
grolu: Yes let's keep this in mind when we refactor the target command. I'm not sure if we still need targetkinds in this case. All targetkinds will then bre represented by a target subcommand and alltargets will be the target root command.

pkg/target/target.go
grolu: Yes, this sounds reasonable. The control plane is always tied to the shoot. Do you think we should discuss this or do we just keep this in mind as an optimization for the future? Changing the target file spec should not have much impact on the user as it is mostly sed for one session only.

pkg/cmd/target/unset.go
```diff
@@ -65,11 +63,12 @@ func runCmdUnset(f util.Factory, o *UnsetOptions) error {
...
 	case TargetKindControlPlane:
 		err = manager.UnsetTargetControlPlane()
-		if err == nil {
-			var currentTarget target.Target
-			currentTarget, err = manager.CurrentTarget()
-			targetName = currentTarget.ShootName()
+		currentTarget, targetErr := manager.CurrentTarget()
+
+		if targetErr != nil {
+			return err
```
holgerkoser, line 69: I don't understand what's going on here. You have to explain this.

holgerkoser (approved):
/ltgm

petersutter (approved):
/lgtm

PR #61, opened 2022-01-11
Configure the kubernetes CLI (kubectl) like the cloud provider CLIs (gcloud, aws, azure, ...) 

holgerkoser:
This PR enhances `gardenctl` with a new command:
```
gardenctl kubectl-env [bash, zsh, fish, powershell]
```
This command generates a script that points KUBECONFIG to the currently targeted cluster for the specified shell. The command is analog to the already existing `cloud-env` command.

Fixes #

**Release note**:

```feature user

```


pkg/cmd/env/template.go
```diff
@@ -0,0 +1,120 @@
...
+	if len(list) > 0 {
+		embedExist = true
+
+		_, err := t.ParseFS(fsys, pattern)
+		if err != nil {
+			return fmt.Errorf("parsing embedded template %q failed: %w", name, err)
+		}
+	}
+
+	if filepath.IsAbs(filename) {
```
holgerkoser, line 106: If no embedded exists and no custom template throw an error.

pkg/cmd/env/options.go
```diff
@@ -0,0 +1,281 @@
...
+
+// Complete adapts from the command line args to the data required.
+func (o *options) Complete(f util.Factory, cmd *cobra.Command, args []string) error {
+	o.Shell = cmd.Name()
+	o.CmdPath = cmd.Parent().CommandPath()
+	o.GardenDir = f.GardenHomeDir()
+
+	switch o.ProviderType {
+	case "kubernetes":
+		o.Template = newTemplate("usage-hint", "templates/kubernetes.tmpl")
```
holgerkoser, line 55: allow override

pkg/target/manager.go
```diff
@@ -62,10 +66,14 @@ type Manager interface {
 	// of a pattern
 	TargetMatchPattern(ctx context.Context, value string) error
 
+	// Kubeconfig returns the kubeconfig for the given target cluster
+	Kubeconfig(ctx context.Context, t Target) ([]byte, error)
+	// WriteKubeconfig creates a kubeconfig file in the temporary directory of the os
+	WriteKubeconfig(data []byte) (string, error)
 	// SeedClient controller-runtime client for accessing the configured seed cluster
 	SeedClient(ctx context.Context, t Target) (client.Client, error)
 	// ShootClusterClient controller-runtime client for accessing the configured shoot cluster
```
petersutter, line 75:
```suggestion
	// ShootClient controller-runtime client for accessing the configured shoot cluster
```

pkg/target/manager.go
```diff
@@ -338,56 +346,91 @@ func (m *managerImpl) patchTargetWithTarget(target Target) error {
...
+		return m.getKubeconfig(t, func(client gardenclient.Client) ([]byte, error) {
+			var namespace string
+
+			if t.ProjectName() != "" {
+				project, err := client.GetProject(ctx, t.ProjectName())
+				if err != nil {
+					return nil, err
+				}
+
+				namespace = *project.Spec.Namespace
```
petersutter, line 360: check that Namespace is not nil

pkg/target/manager.go
```diff
@@ -338,56 +346,91 @@ func (m *managerImpl) patchTargetWithTarget(target Target) error {
...
+		err := os.Mkdir(tempDir, 0700)
+		if err != nil {
+			return "", fmt.Errorf("failed to create temporary gardenctl directory: %w", err)
+		}
 	}
 
-	gardenClient, err := m.GardenClient(t.GardenName())
-	if err != nil {
-		return nil, fmt.Errorf("failed to create garden cluster client: %w", err)
+	filename := filepath.Join(tempDir, fmt.Sprintf("kubeconfig.%x.yaml", sha256.Sum256(data)))
```
petersutter, line 398: as discussed, revert to previous logic (no hash in filename)

pkg/target/manager.go
```diff
@@ -403,46 +446,41 @@ func (m *managerImpl) ShootClusterClient(ctx context.Context, t Target) (client.
 	// the shoot, which is also needed because we need to locate the kubeconfig
 	// on disk.
```
petersutter, line 447:
```go
	// [...] which is also needed because we need to locate the kubeconfig
	// on disk.
```
is this still true?

pkg/cmd/env/template.go
holgerkoser: ✔️

pkg/cmd/env/options.go
holgerkoser: ✔️

.golangci.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - "zz_generated.*"
+linters:
+  enable:
+    - golint
+    - whitespace
+    - wsl
+    - goimports
+linters-settings:
+  goimports:
+    local-prefixes: github.com/gardener
```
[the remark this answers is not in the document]
holgerkoser: ✔️

internal/gardenclient/client.go
```diff
@@ -216,38 +223,98 @@ func (g *clientImpl) GetNamespace(ctx context.Context, name string) (*corev1.Nam
...
+		if !apierrors.IsNotFound(err) {
+			return nil, fmt.Errorf("failed to get kubeconfig for shoot %v: %w", key, err)
+		}
+
+		secret, err := g.GetSecret(ctx, namespace, name+".kubeconfig")
+		if err != nil {
+			return nil, fmt.Errorf("failed to get kubeconfig for shoot %v: %w", key, err)
+		}
+
+		data = secret.Data
```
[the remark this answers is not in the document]
holgerkoser: Good catch. I will remove it.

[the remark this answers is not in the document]
holgerkoser: ✔️

.golangci.yaml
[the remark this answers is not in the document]
holgerkoser: Removed the flags except `--verbose`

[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/cmd/env/env_test.go
```diff
@@ -0,0 +1,85 @@
...
+			for _, c := range subCmds {
+				Expect(cmd.Flag("unset")).To(BeIdenticalTo(flag))
+				s := env.Shell(c.Name())
+				Expect(s).To(BeElementOf(env.ValidShells))
+			}
+		})
+
+	})
+
+	Describe("given a ProviderEnv instance", func() {
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/cmd/env/options.go
```diff
@@ -0,0 +1,281 @@
...
+
+	switch o.ProviderType {
+	case "kubernetes":
+		return o.runKubernetes(f.Context(), manager)
+	default:
+		t := o.CurrentTarget
+		if t.ShootName() == "" {
+			return target.ErrNoShootTargeted
+		} else if t.ProjectName() == "" && t.SeedName() == "" {
+			return target.ErrNeitherProjectNorSeedTargeted
```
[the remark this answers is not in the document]
holgerkoser: Yes. I think, it should be allowed. I think we should also remove the check for both project and seed targeted at same time. Because it will work if the seed is correct otherwise it give an error anyway.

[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
holgerkoser: I have not changed this line but anyway I have fixed it.

pkg/target/manager.go
holgerkoser: ✔️

pkg/target/manager.go
holgerkoser: Implemented the solution we have discussed with the `GCTL_SESSION_ID` which defaults to `ITERM_SESSION_ID` which defaults to a generated v4 uuid.

holgerkoser: ✔️

pkg/target/manager.go
holgerkoser: I don't know exactly but from my point of view it is not true anymore. The origin is the `kubeconfig_cache` which anyway should be removed. The same error is thrown there. I will remove that check here.

PR #63, opened 2022-01-17
[GCP] Do not set IPv6 CIDR on bastion resource

### LINKED ISSUE #32: [GCP] Do not set IPv6 CIDR on bastion resource
GCP has the limitation that you can't set an IPv6 address as `SourceRange` of the firewall rule. If it is set, the `gcp` bastion controller will error.

That's why no IPv6 CIDR should not be set when running `ssh` for a `gcp` shoot https://github.com/gardener/gardenctl-v2/blob/master/pkg/cmd/ssh/command.go#L251-L257

Related discussion https://github.com/gardener/gardener-extension-provider-gcp/pull/288#discussion_r670368282

**Why is this needed**:
To successfully create Bastion resources for `gcp` shoots.

petersutter:
GCP has the limitation that you can't set an IPv6 address as SourceRange of the firewall rule. If it is set, the gcp bastion controller will error.

That's why no IPv6 CIDR should not be set when running ssh for a gcp shoot

Fixes #32 

**Release note**:

```feature user

```


pkg/cmd/ssh/options.go
```diff
@@ -565,6 +565,43 @@ func (o *SSHOptions) Run(f util.Factory) error {
 	return err
 }
 
+func (o *SSHOptions) bastionIngressPolicies(shoot *gardencorev1beta1.Shoot) ([]operationsv1alpha1.BastionIngressPolicy, error) {
```
holgerkoser, line 568: This method only requires the providerType as input parameter. It would keep it simpler as it is an internal method we change change it if more info is needed.

petersutter: ✅

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #64, opened 2022-01-17
Implementation proposal for controlplane kind

grolu:

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/pkg/cmd/target/target.go
+++ b/pkg/cmd/target/target.go
@@ -111,7 +111,7 @@ const (
 	TargetKindSeed         TargetKind = "seed"
 	TargetKindShoot        TargetKind = "shoot"
 	TargetKindPattern      TargetKind = "pattern"
-	TargetKindControlPlane TargetKind = "controlplane"
+	TargetKindControlPlane TargetKind = "control-plane"
 )
 
 var (
--- a/pkg/cmd/target/target.go
+++ b/pkg/cmd/target/target.go
@@ -251,14 +251,15 @@ func (o *TargetOptions) Complete(f util.Factory, cmd *cobra.Command, args []stri
 
 // Validate validates the provided options
 func (o *TargetOptions) Validate() error {
-	if o.Kind == TargetKindControlPlane {
-		// control plane flag can be used without other arguments
-		return nil
-	}
-
-	// reject flag/arg-less invocations
-	if o.Kind == "" || o.TargetName == "" {
-		return errors.New("no target specified")
+	switch o.Kind {
+	case "":
+		return errors.New("no target kind specified")
+	case TargetKindControlPlane:
+		// valid
+	default:
+		if o.TargetName == "" {
+			return fmt.Errorf("target kind %q requires a name argument", o.Kind)
+		}
 	}
 
 	if err := ValidateKind(o.Kind); err != nil {
--- a/pkg/cmd/target/target.go
+++ b/pkg/cmd/target/target.go
@@ -73,11 +73,8 @@ func runCmdTarget(f util.Factory, o *TargetOptions) error {
 		err = manager.TargetShoot(ctx, o.TargetName)
 	case TargetKindPattern:
 		err = manager.TargetMatchPattern(ctx, o.TargetName)
-	default:
-		if o.TargetName == "" && manager.TargetFlags().ControlPlane() {
-			// Allow to target shoot control plane without specifying other arguments
-			err = manager.TargetControlPlane(ctx)
-		}
+	case TargetKindControlPlane:
+		err = manager.TargetControlPlane(ctx)
 	}
 
 	if err != nil {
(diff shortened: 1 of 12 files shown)


pkg/cmd/target/target.go
```diff
@@ -90,14 +87,12 @@ func runCmdTarget(f util.Factory, o *TargetOptions) error {
 	}
 
 	if o.Output == "" {
-		if o.Kind != "" {
+		if o.Kind == TargetKindControlPlane {
+			fmt.Fprintf(o.IOStreams.Out, "Successfully targeted control plane for shoot %q\n", currentTarget.ShootName())
```
petersutter, line 91: maybe
```suggestion
			fmt.Fprintf(o.IOStreams.Out, "Successfully targeted control plane of shoot %q\n", currentTarget.ShootName())
```

pkg/cmd/target/target.go
```diff
@@ -249,14 +246,12 @@ func (o *TargetOptions) Complete(f util.Factory, cmd *cobra.Command, args []stri
...
 
-	o.ControlPlane = tf.ControlPlane()
-
 	return nil
 }
 
 // Validate validates the provided options
 func (o *TargetOptions) Validate() error {
-	if o.ControlPlane {
+	if o.Kind == TargetKindControlPlane {
```
petersutter, line 254: let's discuss about the change in the validate function (in general)

pkg/target/manager.go
```diff
@@ -65,6 +66,8 @@ type Manager interface {
 	UnsetTargetSeed() (string, error)
 	// UnsetTargetShoot unsets the garden shoot configuration
 	UnsetTargetShoot() (string, error)
+	// UnsetTargetControlPlane unsets the garden shoot configuration
```
petersutter, line 69: the go-doc does not fit very vell, or not?

pkg/cmd/target/target.go
holgerkoser: I agree. I propose:
```go
switch o.Kind {
case TargetKindControlPlane:
  break
case "":
  return error.New("error")  
default: 
  if o.TargetName == "" {
    return error.New("error")  
  } 
}
```

pkg/cmd/target/unset_test.go
```diff
@@ -211,6 +187,35 @@ var _ = Describe("Command", func() {
...
+
+		fakeGardenClient := fake.NewClientBuilder().WithObjects(project, shoot).Build()
+
+		// setup command
+		targetProvider := internalfake.NewFakeTargetProvider(currentTarget)
+		clientProvider := internalfake.NewFakeClientProvider()
+		clientProvider.WithClient(gardenKubeconfig, fakeGardenClient)
+
+		factory := internalfake.NewFakeFactory(cfg, nil, clientProvider, nil, targetProvider)
+		cmd := cmdtarget.NewCmdUnset(factory, cmdtarget.NewUnsetOptions(streams))
```
holgerkoser, line 205: This code is repeated in every single test. It should be moved into a `JustBeforeEach` hook. Maybe not as part of this PR but in general.

pkg/target/manager_test.go
```diff
@@ -347,6 +347,22 @@ var _ = Describe("Manager", func() {
 		assertTargetProvider(targetProvider, target.NewTarget(gardenName, prod1Project.Name, "", ""))
 	})
 
+	It("should be able to target control plane for a shoot", func() {
+		t := target.NewTarget(gardenName, prod1Project.Name, "", prod1GoldenShoot.Name)
+		manager, targetProvider := createTestManager(t, *cfg, clientProvider, kubeconfigCache)
+
+		Expect(manager.TargetControlPlane(ctx)).To(Succeed())
+		assertTargetProvider(targetProvider, target.NewTarget(gardenName, prod1Project.Name, "", prod1GoldenShoot.Name).WithControlPlane(true))
```
holgerkoser, line 355:
```suggestion
		assertTargetProvider(targetProvider, t.WithControlPlane(true))
```

pkg/target/manager_test.go
```diff
@@ -462,6 +478,24 @@ var _ = Describe("Manager", func() {
 		assertTargetProvider(targetProvider, t)
 	})
 
+	It("should be able to unset selected control plane", func() {
+		t := target.NewTarget(gardenName, prod1Project.Name, "", prod1AmbiguousShoot.Name).WithControlPlane(true)
+		manager, targetProvider := createTestManager(t, *cfg, clientProvider, kubeconfigCache)
+
+		Expect(manager.UnsetTargetControlPlane()).Should(Equal(prod1AmbiguousShoot.Name))
+		assertTargetProvider(targetProvider, target.NewTarget(gardenName, prod1Project.Name, "", prod1AmbiguousShoot.Name))
```
holgerkoser, line 486:
```suggestion
		assertTargetProvider(targetProvider, t..WithControlPlane(true))
```

pkg/target/manager_test.go
```diff
@@ -462,6 +478,24 @@ var _ = Describe("Manager", func() {
...
+	})
+
+	It("should fail if no control plane targeted", func() {
```
holgerkoser, line 489: Shouldn't this be idempotent and not return an error?

pkg/cmd/target/unset_test.go
grolu: Yes maybe you saw that I already cleaned up this file, also all the test data has been redeclared in every single test. I moved this in the before each function, however skipped the other lines (for now) as I did not want to change the implementation itself. Also, this pattern is repeated in other files e.g. target_test.go. Let's do this in a separate PR.

pkg/target/manager_test.go
grolu: All other unset operations return an error in this case. E.g. `gctl target unset shoot` also fails if no shoot is set. We would need to change it for these cases, too.

holgerkoser: Ok. I would not return error in all these cases but more important is a consistent behavior.

pkg/cmd/target/unset_test.go
holgerkoser: 🆗

holgerkoser (approved):
/lgtm

PR #58, opened 2021-12-14
Added config command

grolu:
**What this PR does / why we need it**:
Added config command to add, configure or delete gardens from the configuration
View the garden configuration
Define context for kubeconfig
Moved patterns from global to gardens in configuration
Dropped support for garden capture group (not required anymore, as each garden has own patterns now)

**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:

**Release note**:

```feature user

```

PR #65, opened 2022-01-20
Config command proposal

holgerkoser:
This PR is an alternative version of the #58 which has the following changes:
- Added config command to add, configure or delete gardens from the configuration
- View the garden configuration
- Define context for kubeconfig
- Moved patterns from global to gardens in configuration
- Dropped support for garden capture group (not required anymore, as each garden has own patterns now)

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/pkg/cmd/config/config.go
+++ b/pkg/cmd/config/config.go
@@ -13,7 +13,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/gardener/gardenctl-v2/internal/util"
-	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
 	"github.com/gardener/gardenctl-v2/pkg/config"
 )
 
@@ -62,82 +61,4 @@ func getConfiguration(f util.Factory) (*config.Config, error) {
 	}
 
 	return config, nil
-}
-
-// NewCmdConfigView returns a new (config) view command.
-func NewCmdConfigView(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
-	use := "view"
-	o := &options{
-		Options: base.Options{
-			IOStreams: ioStreams,
-		},
-		Command: use,
-	}
-	cmd := &cobra.Command{
-		Use:   use,
-		Short: "Print the gardenctl configuration",
-		RunE:  base.WrapRunE(o, f),
-	}
-
-	o.AddFlags(cmd.Flags())
-
-	return cmd
-}
-
-// NewCmdConfigSetGarden returns a new (config) set-garden command.
-func NewCmdConfigSetGarden(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
-	use := "set-garden"
-	o := &options{
-		Options: base.Options{
-			IOStreams: ioStreams,
-		},
-		Command: use,
-	}
-	cmd := &cobra.Command{
-		Use:   use,
-		Short: "modify or add Garden to gardenctl configuration",
-		Long:  "Modify or add Garden to gardenctl configuration. E.g. \"gardenctl config set-garden my-garden --kubeconfig ~/.kube/kubeconfig.yaml\" to configure or add a garden with identity 'my-garden'",
-		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-			suggestions, err := validGardenArgsFunction(f, args)
-			if err != nil {
-				fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
-				return nil, cobra.ShellCompDirectiveNoFileComp
-			}
-
-			return util.FilterStringsByPrefix(toComplete, suggestions), cobra.ShellCompDirectiveNoFileComp
-		},
-		RunE: base.WrapRunE(o, f),
-	}
-
-	o.AddFlags(cmd.Flags())
-
-	return cmd
-}
-
-// NewCmdConfigDeleteGarden returns a new (config) delete-garden command.
-func NewCmdConfigDeleteGarden(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
-	use := "delete-garden"
-	o := &options{
-		Options: base.Options{
-			IOStreams: ioStreams,
-		},
-		Command: use,
-	}
-	cmd := &cobra.Command{
-		Use:   use,
-		Short: "delete Garden from gardenctl configuration",
-		Long:  "Delete Garden from gardenctl configuration. E.g. \"gardenctl config delete-garden my-garden\" to delete my-garden",
-		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-			suggestions, err := validGardenArgsFunction(f, args)
-			if err != nil {
-				fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
-				return nil, cobra.ShellCompDirectiveNoFileComp
-			}
-
-			return util.FilterStringsByPrefix(toComplete, suggestions), cobra.ShellCompDirectiveNoFileComp
-		},
-		RunE: base.WrapRunE(o, f),
-	}
-
-	return cmd
 }
--- a/pkg/cmd/config/config.go
+++ b/pkg/cmd/config/config.go
@@ -36,17 +36,22 @@ The loading order follows these rules:
 	return cmd
 }
 
-func validGardenArgsFunction(f util.Factory, args []string) ([]string, error) {
-	if len(args) > 0 {
-		return nil, nil
-	}
+type cobraValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
 
-	config, err := getConfiguration(f)
-	if err != nil {
-		return nil, err
-	}
+func validGardenArgsFunctionWrapper(f util.Factory, ioStreams util.IOStreams) cobraValidArgsFunction {
+	return func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) > 0 {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
 
-	return config.GardenNames(), nil
+		config, err := getConfiguration(f)
+		if err != nil {
+			fmt.Fprintln(ioStreams.ErrOut, err.Error())
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+
+		return util.FilterStringsByPrefix(toComplete, config.GardenNames()), cobra.ShellCompDirectiveNoFileComp
+	}
 }
 
 func getConfiguration(f util.Factory) (*config.Config, error) {
--- a/pkg/cmd/config/config.go
+++ b/pkg/cmd/config/config.go
@@ -1,0 +1,133 @@
+/*
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
+
+	"github.com/gardener/gardenctl-v2/internal/util"
+
+	"github.com/spf13/cobra"
+)
+
+// NewCmdConfig returns a new config command.
+func NewCmdConfig(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "config",
+		Short: "modify gardenctl configuration file using subcommands",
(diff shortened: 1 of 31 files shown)


pkg/target/manager_test.go
```diff
@@ -313,11 +313,11 @@ var _ = Describe("Target Manager", func() {
 		assertTargetProvider(targetProvider, target.NewTarget(gardenName, prod1Project.Name, "", prod1GoldenShoot.Name))
 	})
 
-	It("should fail to target shoot by matching a pattern if garden is not set", func() {
+	It("should be able to target shoot by matching a pattern if garden is not set", func() {
```
holgerkoser, line 316: This test has been deleted in the original PR. I am not sure why.

pkg/cmd/config/config.go
```diff
@@ -0,0 +1,133 @@
...
+	config := manager.Configuration()
+	if config == nil {
+		return nil, errors.New("failed to get configuration")
+	}
+
+	return config.GardenNames(), nil
+}
+
+// NewCmdConfigView returns a new (config) view command.
+func NewCmdConfigView(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
```
grolu, line 58: I agree with @petersutter - we always use dedicated files for subcommands. Yes, it adds a bit of additional boilerplate code, but keeps functionality separated.

pkg/cmd/config/options.go
```diff
@@ -0,0 +1,174 @@
...
+	case "view":
+		return o.Options.Validate()
+	}
+
+	return nil
+}
+
+// AddFlags adds flags to adjust the output to a cobra command
+func (o *options) AddFlags(flags *pflag.FlagSet) {
+	switch o.Command {
```
grolu, line 87: see my comment above. I'm also a bit unsure if this is a good idea in this case. However there may be cases (e.g. when we refactor the target command) where these approach can help to make the code cleaner, because we have a lot of overlap of the subcommands. But not in this case.

pkg/target/manager_test.go
grolu: I think it has been removed because the test was about the requirement for a garden match in the string to identify the garden if it is not set. However, now with the patterns moved to the garden struct itself, there is no garden capture group anymore and the garden is identified by the regex itself (e.g. regex contains unique garden identifier). Hence, this test is not applicable anymore. Tests for the case that the pattern does not match are still in place.

pkg/cmd/config/options.go
```diff
@@ -0,0 +1,174 @@
...
+	case "delete-garden":
+		return o.runDeleteGarden()
+	case "view":
+		return o.runView()
+	}
+
+	return nil
+}
+
+func (o *options) runSetGarden() error {
```
grolu, line 118: I understand why you moved the code here. As discussed we do not want to have the StringFlags in the config struct. However, what I do not like is the fact that for delete garden there is a function on the struct and for set garden all the logic is here. Maybe we could still have it in the struct. If we do not want to pass the StringFlag types we could also pass string pointers, where nil stands for value not provided. But that is just my personal opinion.

pkg/config/config.go
```diff
@@ -173,8 +174,38 @@ const (
...
+			}
+
+			patternMatch = match
+		}
+	}
+
+	if patternMatch != nil {
+		// Did not match pattern of current garden, but did match other pattern
+		return patternMatch, nil
+	}
```
petersutter, line 201: let's quickly talk about this logic for my understanding

pkg/config/config.go
```diff
@@ -173,8 +174,38 @@ const (
...
+
+	return nil, errors.New("the provided value does not match any pattern")
+}
+
+// matchPattern matches pattern with provided list of patterns
```
petersutter, line 206: maybe it makes sense to add it in the go-doc that no error is returned in case the pattern did not match?

pkg/target/manager.go
```diff
@@ -143,8 +143,11 @@ func (m *managerImpl) Configuration() *config.Config {
 	return m.config
 }
 
-func (m *managerImpl) TargetGarden(ctx context.Context, gardenNameOrAlias string) error {
-	tb := NewTargetBuilder(m.config, m.clientProvider)
+func (m *managerImpl) TargetGarden(ctx context.Context, identity string) error {
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target garden: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -167,24 +170,26 @@ func (m *managerImpl) UnsetTargetGarden() (string, error) {
...
+
+		return nil
+	})
 }
 
 func (m *managerImpl) TargetProject(ctx context.Context, projectName string) error {
-	tb := NewTargetBuilder(m.config, m.clientProvider)
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target project: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -208,21 +213,24 @@ func (m *managerImpl) UnsetTargetProject() (string, error) {
...
+
+		return nil
+	})
 }
 
 func (m *managerImpl) TargetSeed(ctx context.Context, seedName string) error {
-	tb := NewTargetBuilder(m.config, m.clientProvider)
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target seed: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -246,19 +254,22 @@ func (m *managerImpl) UnsetTargetSeed() (string, error) {
...
+
+		return nil
+	})
 }
 
 func (m *managerImpl) TargetShoot(ctx context.Context, shootName string) error {
-	tb := NewTargetBuilder(m.config, m.clientProvider)
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target shoot: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -282,20 +293,23 @@ func (m *managerImpl) UnsetTargetShoot() (string, error) {
...
+
+		return nil
+	})
 }
 
 func (m *managerImpl) TargetControlPlane(ctx context.Context) error {
-	tb := NewTargetBuilder(m.config, m.clientProvider)
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target shoot control plane: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -330,16 +344,25 @@ func (m *managerImpl) UnsetTargetControlPlane() error {
...
+
+	tm, err := m.config.MatchPattern(value, gardenName)
 	if err != nil {
-		return fmt.Errorf("failed to get current target: %w", err)
+		return fmt.Errorf("error occurred while trying to match value: %w", err)
+	}
+
+	tb, err := NewTargetBuilder(m.config, m.clientProvider)
+	if err != nil {
+		return fmt.Errorf("failed to target with match pattern: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -153,9 +156,9 @@ func (m *managerImpl) TargetGarden(ctx context.Context, gardenNameOrAlias string
 
 	tb.Init(currentTarget)
 
-	target, err := tb.SetGarden(gardenNameOrAlias).Build()
+	target, err := tb.SetGarden(identity).Build()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to target garden: %w", err)
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/target/manager.go
```diff
@@ -167,24 +170,26 @@ func (m *managerImpl) UnsetTargetGarden() (string, error) {
...
+	if targetedGarden == "" {
+		return "", ErrNoGardenTargeted
 	}
 
-	return "", ErrNoGardenTargeted
+	return targetedGarden, m.patchTarget(func(t *targetImpl) error {
+		t.Garden = ""
+		t.Project = ""
+		t.Seed = ""
+		t.Shoot = ""
```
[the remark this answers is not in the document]
holgerkoser: ✔️

pkg/config/config.go
holgerkoser: ✔️

pkg/cmd/config/options.go
holgerkoser: If we don't that somebody operates on the struct data we have to hide it behind an interface. As of now we should consider the method of the Config struct as utility methods which could be used or not. To make it consistent i have moved the delete garden impl from the config struct to the command.

pkg/cmd/config/options.go
holgerkoser: I have split all subcommands into single files.

pkg/cmd/config/config.go
holgerkoser: I have split all subcommands into their own single file.

pkg/target/manager_test.go
holgerkoser: All the tests always succeed because the target was passed by reference to the fakeTargetProvider (see https://github.com/gardener/gardenctl-v2/pull/65/commits/5637c84e823fedf8f5e173369543bf98acc680b8)

grolu (approved):
/lgtm

pkg/cmd/config/set_garden.go
```diff
@@ -0,0 +1,166 @@
...
+// AddFlags adds flags to adjust the output to a cobra command
+func (o *setGardenOptions) AddFlags(flags *pflag.FlagSet) {
+	flags.Var(&o.KubeconfigFlag, "kubeconfig", "path to kubeconfig file for this Garden cluster")
+	flags.Var(&o.ContextFlag, "context", "override the current-context of the garden cluster kubeconfig")
+	flags.StringArrayVar(&o.Patterns, "pattern", nil, `define regex match patterns for this garden for custom input formats for targeting.
+Use named capturing groups to match target values.
+Supported capturing groups: project, namespace, shoot.
+Note that if you set this flag it will overwrite the pattern list in the config file.
+You may specify any number of extra patterns.
+Example: ^((?Pmy-garden[^/]+)/)?shoot--(?P<project>.+)--(?P<shoot>.+)$`)
```
holgerkoser, line 100: Invalid regular expression in the example.

pkg/cmd/config/set_garden.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to configure garden: %w", err)
+	}
+
+	fmt.Fprintf(o.IOStreams.Out, "Successfully configured garden %q\n", o.Name)
+
+	return nil
+}
+
+func validatePatterns(patterns []string) error {
```
holgerkoser, line 142: How should this flag work in detail. Should it be possible to delete a single or all patterns?

pkg/cmd/config/set_garden.go
```diff
@@ -0,0 +1,166 @@
...
+		if o.ContextFlag.Provided() {
+			garden.Context = o.ContextFlag.Value()
+		}
+
+		if o.Patterns != nil {
+			firstPattern := o.Patterns[0]
+			if len(firstPattern) > 0 {
+				garden.Patterns = o.Patterns
+			} else {
+				garden.Patterns = []string{}
```
holgerkoser, line 120: If there is no pattern it should be `nil`.
```suggestion
				garden.Patterns =nil
```

pkg/cmd/config/set_garden.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func validatePatterns(patterns []string) error {
+	if patterns == nil || patterns[0] == "" && len(patterns) == 1 {
+		return nil
+	}
+
+	for i, p := range patterns {
+		if p == "" {
+			return fmt.Errorf("pattern[%d] must not be empty", i)
```
holgerkoser, line 149: Woe should the pattern with the error referenced in the error message?

pkg/config/config.go
holgerkoser: @grolu will create an own PR for this.

pkg/cmd/config/set_garden.go
holgerkoser: ✔️

pkg/cmd/config/set_garden.go
holgerkoser: ✔️

PR #66, opened 2022-01-21
Ensure valid target when targeting control plane

grolu:
Ensure valid target is returned by the target builder

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/pkg/target/target_builder.go
+++ b/pkg/target/target_builder.go
@@ -164,32 +164,7 @@ func (b *targetBuilderImpl) SetShoot(ctx context.Context, name string) TargetBui
 			return ErrNoGardenTargeted
 		}
 
-		gardenClient, err := b.getGardenClient(t.GardenName())
-		if err != nil {
-			return err
-		}
-
-		shoot, err := gardenClient.FindShoot(ctx, t.WithShootName(name).AsListOption())
-		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/target/target_builder.go
```diff
@@ -164,49 +164,58 @@ func (b *targetBuilderImpl) SetShoot(ctx context.Context, name string) TargetBui
...
+	if err != nil {
+		return err
+	}
+
+	shoot, err := gardenClient.FindShoot(ctx, t.WithShootName(name).AsListOption())
+	if err != nil {
+		return fmt.Errorf("failed to fetch shoot: %w", err)
+	}
+
+	if t.ProjectName() == "" {
```
holgerkoser, line 203:
```suggestion
	if t.Project == "" {
```

holgerkoser: Maybe also for GardenName because we need to use targetImpl here.

holgerkoser (approved):
Apart from the cosmetic suggestion to not use the getter of the interface but directly the property on the struct, the PR looks very good to me. 
/lgtm

petersutter (approved):
/lgtm

pkg/target/target_builder.go
grolu: Yes, I was wondering the same - so let's always use the property I will change it

PR #68, opened 2022-01-24
Proposal for improving MatchPattern

grolu:
Refactored function and added tests

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -101,7 +101,7 @@ func (config *Config) IndexOfGarden(name string) (int, bool) {
 	return -1, false
 }
 
-// Garden returns a Garden cluster from the list of configured Gardens
+// GardenNames returns a slice containing the names of the configured Gardens
 func (config *Config) GardenNames() []string {
 	names := []string{}
 	for _, g := range config.Gardens {
@@ -182,7 +182,7 @@ func (config *Config) MatchPattern(gardenName string, value string) (*PatternMat
 		}
 	}
 
-	var matches []*PatternMatch
+	var patternMatch *PatternMatch
 
 	for _, g := range config.Gardens {
 		match, err := matchPattern(g.Patterns, value)
@@ -192,19 +192,20 @@ func (config *Config) MatchPattern(gardenName string, value string) (*PatternMat
 
 		if match != nil {
 			match.Garden = g.Name
-			matches = append(matches, match)
+
+			if patternMatch != nil {
+				return nil, errors.New("the provided value resulted in an ambiguous match")
+			}
+
+			patternMatch = match
 		}
 	}
 
-	if len(matches) == 1 {
-		return matches[0], nil
+	if patternMatch == nil {
+		return nil, errors.New("the provided value does not match any pattern")
 	}
 
-	if len(matches) > 1 {
-		return nil, errors.New("the provided value resulted in an ambiguous match")
-	}
-
-	return nil, errors.New("the provided value does not match any pattern")
+	return patternMatch, nil
 }
 
 // matchPattern matches pattern with provided list of patterns
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -164,8 +164,25 @@ const (
 
 // MatchPattern matches a string against patterns defined in gardenctl config
 // If matched, the function creates and returns a PatternMatch from the provided target string
-func (config *Config) MatchPattern(value string, gardenName string) (*PatternMatch, error) {
-	var patternMatch *PatternMatch
+func (config *Config) MatchPattern(gardenName string, value string) (*PatternMatch, error) {
+	if gardenName != "" {
+		g, err := config.Garden(gardenName)
+		if err != nil {
+			return nil, err
+		}
+
+		match, err := matchPattern(g.Patterns, value)
+		if err != nil {
+			return nil, err
+		}
+
+		if match != nil {
+			match.Garden = g.Name
+			return match, nil
+		}
+	}
+
+	var matches []*PatternMatch
 
 	for _, g := range config.Gardens {
 		match, err := matchPattern(g.Patterns, value)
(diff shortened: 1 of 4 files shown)


pkg/config/config.go
```diff
@@ -164,30 +164,44 @@ const (
...
+		if match != nil {
+			match.Garden = g.Name
+			matches = append(matches, match)
 		}
 	}
 
-	if patternMatch != nil {
-		// Did not match pattern of current garden, but did match other pattern
-		return patternMatch, nil
+	if len(matches) == 1 {
```
holgerkoser, line 199: We could early exit from loop and I would check first for errors and return successfully at the end.  
```go
        var patternMatch *PatternMatch
        
	for _, g := range config.Gardens {
		match, err := matchPattern(g.Patterns, value)
		if err != nil {
			return nil, err
		}

		if match != nil {
			match.Garden = g.Name
                         if patternMatch != nil {
                         	return nil, errors.New("the provided value resulted in an ambiguous match")
                         }
                         
                         patternMatch = match
		}
	}
	
	if patternMatch == nil {
	         return nil, errors.New("the provided value does not match any pattern")
	}
	
	return matches[0], nil
```

petersutter (approved):
/lgtm 

However I also like holgers suggestion, even more simplified

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and clusterIdentity1 is set in current target (non unique case / keep garden)",
+			clusterIdentity1,
+			"",
+			clusterIdentity1),
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and clusterIdentity2 is set in current target (non unique case / keep garden)",
+			clusterIdentity2,
+			"",
+			clusterIdentity2),
```
petersutter, line 88: is there a difference between these two? or could you get rid of one of them?

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+	var patternValue = func(prefix string) string {
+		value := fmt.Sprintf("shoot--%s--%s", project, shoot)
+
+		if prefix != "" {
+			value = fmt.Sprintf("%s/%s", prefix, value)
+		}
+		return value
+	}
+
+	DescribeTable("Validation of valid target values",
```
holgerkoser, line 56:
```suggestion
	DescribeTable("MatchPattern returns a match",
```

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+		func(currentGardenName string, patternPrefix string, expectedGarden string) {
+			value := patternValue(patternPrefix)
+			expectedPM := &config.PatternMatch{Garden: expectedGarden, Project: project, Shoot: shoot}
+			match, err := cfg.MatchPattern(currentGardenName, value)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(match).To(Equal(expectedPM))
+		},
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and garden is set in current target (unique case / keep garden)",
```
holgerkoser, line 65:
```suggestion
			"when the value contains a gardenName it's value is correctly extracted",
```

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+			clusterIdentity2,
+			clusterIdentity1,
+			clusterIdentity1),
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and clusterIdentity1 is set in current target (non unique case / keep garden)",
+			clusterIdentity1,
+			"",
+			clusterIdentity1),
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and clusterIdentity2 is set in current target (non unique case / keep garden)",
```
holgerkoser, line 85:
```suggestion
			"when the targetValue does not contain a gardenName the given gardenName is taken",
```

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+			"",
+			clusterIdentity1,
+			clusterIdentity1),
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and other garden is set in current target (unique case / switch garden)",
+			clusterIdentity2,
+			clusterIdentity1,
+			clusterIdentity1),
+		Entry(
+			"should succeed if pattern for value is (potential) non unique pattern and clusterIdentity1 is set in current target (non unique case / keep garden)",
```
holgerkoser, line 80: What is the difference between the last two tests?

pkg/config/config_test.go
holgerkoser: Whats ist the difference between the first three tests?

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+
+	DescribeTable("Validation of invalid target values",
+		func(currentGardenName string, patternPrefix string, expectedErrorString string) {
+			value := patternValue(patternPrefix)
+			_, err := cfg.MatchPattern(currentGardenName, value)
+			Expect(err).To(HaveOccurred())
+			Expect(err).To(MatchError(ContainSubstring(expectedErrorString)))
+		},
+		Entry(
+			"should fail if pattern for value is (potential) non unique pattern and garden is not set in current target (non unique case / ambiguous match / fail to determine garden)",
```
holgerkoser, line 99:
```suggestion
			"when no garden is given and the value matches multiple gardens",
```

pkg/config/config_test.go
```diff
@@ -31,20 +31,87 @@ var _ = Describe("Config", func() {
...
+			"",
+			"",
+			"the provided value resulted in an ambiguous match"),
+		Entry(
+			"should fail if pattern for value is not found and garden is set",
```
holgerkoser, line 104:
```suggestion
			"when a garden is given and the value does not match any pattern",
```

pkg/config/config_test.go
grolu: It is important to not change the cluster. I wanted to cover this by testing with both clusters. Depending on the code, one of those two tests could fail but it I cannot for-see which one (as it depends on the code)

pkg/config/config_test.go
```diff
@@ -31,20 +32,92 @@ var _ = Describe("Config", func() {
...
+			clusterIdentity2,
+			"",
+			clusterIdentity2),
+	)
+
+	DescribeTable("MatchPattern returns an error",
+		func(currentGardenName string, patternPrefix string, expectedErrorString string) {
+			value := patternValue(patternPrefix)
+			_, err := cfg.MatchPattern(currentGardenName, value)
+			Expect(err).To(HaveOccurred())
```
holgerkoser, line 96: MatchError already ensure that an error occurs.
```suggestion
```

petersutter (approved):
/lgtm

PR #74, opened 2022-01-27
Session isolation for `provider-env` command

### LINKED ISSUE #73: command `provider-env` should not create any state outside of the temporary session directory
For the cloud providers `alicloud`, `aws`, `hcloud`, `kubernetes` and `openstack` the generated script only set environment variables in order to configure the corresponding cloud provider CLI. But for `azure` and `gcp` we cannot pass the credentials to the cli and we have to call `az login` or `gcloud activate-service-account` to be able to use the CLI. These commands create persistent state in the configuration directory of the CLI which has side effects across shell session. Currently the user has to call the `provider-env` command with the unset options `-u` to reset this state (logout). Instead the generated script should point  `az` and `gcloud` to a configuration directory inside the `gardenctl` session directory. This means the calls to `az login` or `gcloud activate-service-account` will not create state outside of the scope of the current shell session and logout has not to be called.

In the template values https://github.com/gardener/gardenctl-v2/blob/master/pkg/cmd/env/options.go#L184-L202 the following code needs to added:
```go
	switch o.ProviderType {
	case "azure":
		configDir := filepath.join(o.SessionDirectory, ".config", "az")
                 err = os.MkdirAll(configDir, 0700)
                 if err != nil {
			return err
		}

                data["configDir"] = configDir
	case "gcp":
		credentials := make(map[string]interface{})

		serviceaccountJSON, err := parseGCPCredentials(secret, &credentials)
		if err != nil {
			return err
		}

		data["credentials"] = credentials
		data["serviceaccount.json"] = string(serviceaccountJSON)
		configDir := filepath.join(o.SessionDirectory, ".config", "gcloud")
                 err = os.MkdirAll(configDir, 0700)
                 if err != nil {
			return err
		}

                data["configDir"] = configDir
	case "openstack":
		authURL, err := getKeyStoneURL(cloudProfile, shoot.Spec.Region)
		if err != nil {
			return err
		}

		data["authURL"] = authURL
	}
```

**Azure template**
```bash
...
export AZURE_SUBSCRIPTION_ID={{.subscriptionID | shellEscape}};
export AZURE_CONFIG_DIR={{.configDir | shellEscape}};
az login --service-principal --username "$AZURE_CLIENT_ID" --password "$AZURE_CLIENT_SECRET" --tenant "$AZURE_TENANT_ID";
az account set --subscription "$AZURE_SUBSCRIPTION_ID";
``` 
**GCP template**
```bash
...
export CLOUDSDK_COMPUTE_REGION={{.region | shellEscape}};
export CLOUDSDK_CONFIG={{.configDir | shellEscape}};
gcloud auth activate-service-account $GOOGLE_CREDENTIALS_ACCOUNT --key-file <(printf "%s" "$GOOGLE_CREDENTIALS");
```

I think we could remove the unset flag and also remove the logout hint in both cases.

**Why is this needed**:
If the user closes a shell session (e.g by closing the window or entering exit) there should be no leftovers if he opens new shell in a new window. If the user has two shell sessions in two separate windows there should no side effects in the second window if he he runs a `gardenctl` in the first window. Each session should be completely isolated from other sessions.

holgerkoser:
If the user closes a shell session (e.g by closing the window or entering exit) there should be no leftovers if he opens new shell in a new window. If the user has two shell sessions in two separate windows there should be no side effects in the second window if he runs a `gardenctl` in the first window. Each session should be completely isolated from other sessions.

Fixes #73

**Release note**:

```feature user

```

### DIFF
--- a/pkg/cmd/env/options.go
+++ b/pkg/cmd/env/options.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"runtime"
 
@@ -33,6 +34,8 @@ type options struct {
 	Shell string
 	// GardenDir is the configuration directory of gardenctl.
 	GardenDir string
+	// SessionDir is the session directory of gardenctl.
+	SessionDir string
 	// CmdPath is the path of the called command.
 	CmdPath string
 	// CurrentTarget is the current target
@@ -56,6 +59,13 @@ func (o *options) Complete(f util.Factory, cmd *cobra.Command, args []string) er
 			return err
 		}
 	}
+
+	manager, err := f.Manager()
+	if err != nil {
+		return err
+	}
+
+	o.SessionDir = manager.SessionDir()
 
 	return nil
 }
@@ -182,6 +192,13 @@ func execTmpl(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret,
 	}
 
 	switch o.ProviderType {
+	case "azure":
+		configDir, err := createProviderConfigDir(o.SessionDir, o.ProviderType)
+		if err != nil {
+			return err
+		}
+
+		data["configDir"] = configDir
 	case "gcp":
 		credentials := make(map[string]interface{})
 
@@ -190,6 +207,12 @@ func execTmpl(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret,
 			return err
 		}
 
+		configDir, err := createProviderConfigDir(o.SessionDir, o.ProviderType)
+		if err != nil {
+			return err
+		}
+
+		data["configDir"] = configDir
 		data["credentials"] = credentials
 		data["serviceaccount.json"] = string(serviceaccountJSON)
 	case "openstack":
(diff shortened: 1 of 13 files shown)


pkg/cmd/env/options.go
```diff
@@ -50,11 +53,19 @@ func (o *options) Complete(f util.Factory, cmd *cobra.Command, args []string) er
...
 		if err := o.Template.ParseFiles(filename); err != nil {
 			return err
 		}
+	case "azure", "gcp":
+		manager, err := f.Manager()
+		if err != nil {
+			return err
+		}
+
+		o.SessionDir = manager.SessionDir()
```
petersutter, line 68: hm, I'm wondering if it wouldn't be better to split the options for each use case: kubectl-env vs provider-env

sessionDir would only be set on the provider-env options. This would make it clearer when which option is available (without having to clarify it in the go-doc). Also we would get rid of the switches, not sure if there is a lot of code that

If we would agree on that, I'm also fine to postpone this refactoring and do it in a separate PR later on. WDYT?

pkg/cmd/env/templates/azure.tmpl
```diff
@@ -22,11 +24,13 @@ set -e AZURE_CLIENT_ID;
 set -e AZURE_CLIENT_SECRET;
 set -e AZURE_TENANT_ID;
 set -e AZURE_SUBSCRIPTION_ID;
+set -e AZURE_CONFIG_DIR;
```
petersutter, line 27: Maybe I'm too early with my review (PR is still marked as draft) and you planned to do the following anyhow:

> Since these actions are persistent, they should be undone when closing the shell or changing the target using the --unset flag.

Shouldn't the config dir be removed by the unset script?

pkg/cmd/env/options.go
holgerkoser: I agree and will create an issue and make that change in a separate PR.

pkg/cmd/env/templates/azure.tmpl
holgerkoser: ✔️

pkg/cmd/env/options.go
holgerkoser: #76

petersutter (approved):
/lgtm

PR #80, opened 2022-01-31
Do not remove config dir because revoke and logout commands will fails

holgerkoser:
This PR fixes a bug that has been introduced with PR #74. It is not possible to remove cloud provider config dir in the session folder before the generated provider-env unset script is executed because the logout / revoke commands in the script will fail.  

Fixes #

**Release note**:

```feature user

```


petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #86, opened 2022-02-01
Write and symlink kubeconfig on change of target

holgerkoser:
With this PR it is possible to change the target cluster of `kubectl` directly with the `gardenctl target ...` command without having to adjust the `KUBECONFIG` environment variable again and again. Whether the current target cluster of gardenctl and kubectl should be synchronized or not can be set using the configrations property link-kubeconfig or via environment variable `GCTL_LINK_KUBECONFIG`.

#### Behavior without symlink | linkKubeconfig: false
The `target` command only changes the current target cluster of gardenctl, but does not download the corresponding kubeconfig file. This happens only when the `kubectl-env` command is executed. If the user evals the generated script the `KUBECONFIG` environment variable points the downloaded kubeconfig file for the targeted cluster.
Link-Kubeconfig-2
The generated script looks something like this:
```bash
export KUBECONFIG=$TMPDIR/garden/<gctl-session-id>/kubeconfig.<hash>.yaml
```

#### Behavior with symlink | linkKubeconfig: true
The `target` command changes the current target cluster of gardenctl, downloads the corresponding kubeconfig file and creates a stable symlink to the current kubeconfig file in the session directory. If the target is changed the symlink points to the new target but the `KUBECONFIG` environment variable does not need to be changed. The `kubectl-env` command now does nothing except create the script to set the `KUBECONFIG` environment variable. This works even if nothing is targeted and could be done in the shell profile.
Link-Kubeconfig-1
The generated script only needs to be evaluated once per shell session and looks something like this:
```bash
export KUBECONFIG=$TMPDIR/garden/<gctl-session-id>/kubeconfig.yaml
```
#### Test
Test are not yet changed !!!

Fixes #

**Release note**:

```feature user

```


RaphaelVogel:
nice, waiting to get this feature 👍

pkg/cmd/env/options.go
```diff
@@ -140,14 +148,20 @@ func (o *options) runKubernetes(ctx context.Context, manager target.Manager) err
...
-		if err != nil {
-			return err
-		}
+		var filename string
 
-		filename, err := manager.WriteClientConfig(config)
-		if err != nil {
-			return err
+		if o.Symlink {
+			filename = path.Join(o.SessionDir, "kubeconfig.yaml")
```
holgerkoser, line 154: Add error message if target exists but no symlink

pkg/config/config.go
```diff
@@ -56,6 +59,14 @@ func LoadFromFile(filename string) (*Config, error) {
 
 	config := &Config{Filename: filename}
 
+	// We don't want a dependency to root command here
+	str, ok := os.LookupEnv("GCTL_LINK_KUBECONFIG")
+	if ok {
+		if val, err := strconv.ParseBool(str); err == nil {
```
petersutter, line 65: shouldn't we let the user know if `GCTL_LINK_KUBECONFIG` could not be parsed to bool indicating a wrong value or just silently ignore?

pkg/config/config.go
```diff
@@ -21,6 +22,8 @@ import (
 type Config struct {
 	// Filename is the name of the gardenctl configuration file
 	Filename string `yaml:"-"`
+	// LinkKubeconfig defines if kubeconfig is symlinked with the target
+	LinkKubeconfig *bool `yaml:"link-kubeconfig,omitempty"`
```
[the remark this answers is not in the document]
holgerkoser: Do you prefer camelCase? In kubeconfig we have `certificate-authority-data` or `current-context` but I know in gardener we also have camelCase.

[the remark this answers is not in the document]
holgerkoser: I have change it to `linkKubeconfig`

pkg/cmd/env/options.go
holgerkoser: ✔️

petersutter (approved):
/lgtm

PR #91, opened 2022-02-01
fixed capitalization for our commands

grolu:
I tried to fix the places where we have help texts starting with small character.
I'm not really satisfied with the result as there are command texts that we cannot influence like `completion` (starts with small letter) or `help` (starts with capital letter).
Moreover, the flags also sometimes start with small and sometimes with capital letter, even when they come from the same source (e.g. klog flags). I'm a bit lost actually. Let's have a look at this together.

Fixes #

**Release note**:

```feature user

```


holgerkoser (approved):
/lgtm

PR #95, opened 2022-02-03
Enh/target command refactoring

grolu:
Refactored target kinds into subcommands.
This is a first minimally invasive approach to cleanup and provide decent help for each command.
Further cleanup / refactoring is possible but I think it is good for now.
Target Flags should no longer be global but I will do this in a follow-up PR.

Fixes #

**Release note**:

```feature user
Target command now has subcommands for each target kind. This improves working with the command, especially regarding command help
```


pkg/cmd/target/control-plane.go
```diff
@@ -0,0 +1,36 @@
+/*
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Projecter contributors
```
petersutter, line 2: 2022 (also with the other files)

grolu: Yes, we can update all files when we have less PRs open. WDYT?

petersutter: sorry I meant:
2022 (also with the other newly introduced files)

grolu: done

pkg/cmd/target/control-plane.go
```diff
@@ -0,0 +1,36 @@
...
+func NewCmdTargetControlPlane(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
+	o := &TargetOptions{
+		Kind: TargetKindControlPlane,
+		Options: base.Options{
+			IOStreams: ioStreams,
+		},
+	}
+	cmd := &cobra.Command{
+		Use:   "control-plane",
+		Short: "Set control plane for next operations",
```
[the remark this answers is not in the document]
grolu: ok. I updated the help texts

pkg/cmd/target/control-plane.go
```diff
@@ -0,0 +1,36 @@
...
+		Example: `# target control-plane for current shoot cluster
+gardenctl target control-plane`,
```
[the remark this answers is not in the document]
grolu: I added more examples. Do we need even more?

pkg/cmd/target/garden.go
```diff
@@ -0,0 +1,36 @@
...
+		Kind: TargetKindGarden,
+		Options: base.Options{
+			IOStreams: ioStreams,
+		},
+	}
+	cmd := &cobra.Command{
+		Use:   "garden",
+		Short: "Set garden for next operations",
+		Example: `# target garden with name garden_name
+gardenctl target garden garden_name`,
```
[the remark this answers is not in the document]
grolu: done

pkg/cmd/target/target.go
```diff
@@ -18,33 +17,31 @@ import (
...
-				return err
-			}
+# target shoot "my-shoot" in already selected project
+gardenctl target shoot my-shoot
 
-			return runCmdTarget(f, o)
-		},
+# Target shoot control-plane using values that match a pattern defined for a specific landscape
+gardenctl target value/that/matches/pattern --control-plane`,
+		RunE: runCmdTargetWrapper(f, o),
```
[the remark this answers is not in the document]
grolu: The `WrapRunE` function expects the run function implemented on the options struct. As for this change I did not introduce separate options, it is not possible to use this function. We can discuss if we want to have it that way though. Again, the question is to what level we should refactor the command in this PR. If we want to incorporate it in the first release, I would suggest to keep the changes as small as possible.

pkg/cmd/target/target.go
```diff
@@ -252,8 +242,6 @@ func (o *TargetOptions) Complete(f util.Factory, cmd *cobra.Command, args []stri
 // Validate validates the provided options
 func (o *TargetOptions) Validate() error {
 	switch o.Kind {
-	case "":
-		return errors.New("no target kind specified")
```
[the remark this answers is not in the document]
grolu: Yes, I guess so. There is no `kind` anymore that you can specify here. If you use the target command directly, you can either specify everything using flags or you target via pattern or a combination of both. The `target kind` are now subcommands.

pkg/cmd/target/unset.go
```diff
@@ -21,7 +21,12 @@ import (
 func NewCmdUnset(f util.Factory, o *UnsetOptions) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "unset",
-		Short: "Unset target, e.g. \"gardenctl target unset shoot\" to unset currently targeted shoot",
+		Short: "Unset target",
+		Example: `# unset selected shoot
```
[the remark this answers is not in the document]
grolu: We should discuss which term is the better one. Have no strong opinion here.

pkg/cmd/target/target.go
```diff
@@ -54,6 +51,22 @@ func NewCmdTarget(f util.Factory, o *TargetOptions) *cobra.Command {
 	return cmd
 }
 
+type cobraRunFunction = func(cmd *cobra.Command, args []string) error
+
+func runCmdTargetWrapper(f util.Factory, o *TargetOptions) cobraRunFunction {
```
holgerkoser, line 56: Use `base.WrapRunE` instead and implement Run on TargetOptions.
```go
// Run does the actual work of the command.
func (o *Options) Run(util.Factory) error {
	return runCmdTarget(f, o)
}
```

pkg/cmd/target/target.go
```diff
@@ -128,23 +141,21 @@ func ValidateKind(kind TargetKind) error {
...
-			string(TargetKindPattern),
-			string(TargetKindControlPlane),
-		}, nil
-	}
+type cobraValidArgsFunction = func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
 
-	kind := TargetKind(strings.TrimSpace(args[0]))
-	if err := ValidateKind(kind); err != nil {
-		return nil, err
+func validTargetFunctionWrapper(f util.Factory, ioStreams util.IOStreams, kind TargetKind) cobraValidArgsFunction {
```
holgerkoser, line 132: For this PR I would keep it like it is now and not introduce the anonymous func.

pkg/cmd/target/target.go
grolu: done

pkg/cmd/target/control-plane.go
```diff
@@ -23,11 +23,15 @@ func NewCmdTargetControlPlane(f util.Factory, ioStreams util.IOStreams) *cobra.C
...
-		Short: "Set control plane for next operations",
-		Example: `# target control-plane for current shoot cluster
-gardenctl target control-plane`,
+		Short: "Target the control plane of the shoot",
+		Long:  "Target the control plane of the shoot cluster to set the scope for the next operations",
+		Example: `# target control-plane of currently selected shoot cluster
+gardenctl target control-plane
+
+# target control-plane of shoot my-shoot
+gardenctl target control-plane --shoot my-shoot --project my-project --garden my-garden`,
```
petersutter, line 32:
```suggestion
gardenctl target control-plane --garden my-garden --project my-project --shoot my-shoot`,
```
turn it around, otherwise you will have a hard time with the completions

pkg/cmd/target/project.go
```diff
@@ -23,11 +23,15 @@ func NewCmdTargetProject(f util.Factory, ioStreams util.IOStreams) *cobra.Comman
 	}
 	cmd := &cobra.Command{
 		Use:   "project",
-		Short: "Set project for next operations",
-		Example: `# target project with name project_name
-gardenctl target project project_name`,
+		Short: "Target a shoot",
+		Long:  "Target a shoot to set the scope for the next operations",
```
petersutter, line 27:
```suggestion
		Short: "Target a project",
		Long:  "Target a project to set the scope for the next operations",
```

pkg/cmd/target/shoot.go
```diff
@@ -23,11 +23,15 @@ func NewCmdTargetShoot(f util.Factory, ioStreams util.IOStreams) *cobra.Command
...
-		Short: "Set shoot for next operations",
-		Example: `# target shoot with name shoot_name
-gardenctl target shoot shoot_name`,
+		Short: "Target a shoot",
+		Long:  "Target a shoot to set the scope for the next operations",
+		Example: `# target shoot with name my-shoot of currently selected project
+gardenctl target shoot my-shoot
+
+# target shoot with name my-shoot of project my-project
+gardenctl target shoot my-shoot --project my-project --garden my-garden`,
```
petersutter, line 32:
```suggestion
gardenctl target shoot my-shoot --garden my-garden --project my-project`,
```
better for the completion experience would be `gardenctl target shoot --garden my-garden --project my-project my-shoot` but it makes the example harder to read

pkg/cmd/target/target.go
```diff
@@ -18,33 +17,31 @@ import (
...
 
-			if err := o.Validate(); err != nil {
-				return err
-			}
+# target shoot "my-shoot" of currently selected project
+gardenctl target shoot my-shoot
 
-			return runCmdTarget(f, o)
-		},
+# Target shoot control-plane using values that match a pattern defined for a specific landscape
```
petersutter, line 35:
```suggestion
# Target shoot control-plane using values that match a pattern defined for a specific garden
```
as you were replacing `landscape` with garden in one of your latest commits, I guess this one also needs to be replaced

pkg/cmd/target/unset.go
```diff
@@ -21,7 +21,12 @@ import (
...
+gardenctl target unset shoot
+
+# unset garden and everything targeted inside it
```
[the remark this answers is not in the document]
petersutter: Maybe it's better to be more explicit. Like "Unset garden. This will also unset a targeted project, shoot, seed and control plane" or something like that.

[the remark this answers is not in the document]
grolu: ok. makes sense I guess.

pkg/cmd/target/control-plane.go
grolu: ok

pkg/cmd/target/project.go
grolu: nice catch

pkg/cmd/target/shoot.go
grolu: ok let's make the examples always like you would do when making use of completion functionality

grolu: BUT keep the subcommand before the flags as you said, it is much harder to read

pkg/cmd/target/target.go
grolu: Yes, I changed it to garden to not confuse the users. We always use the term garden, yes a garden identifies a landscape but nevertheless I thought it would be better to keep things simple.

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #96, opened 2022-02-08
Use .login suffix for Seed Secrets. Fallback to .oidc

petersutter:
With this PR `gardenctl` will look for a secret named `<seed-name>.login` to get the kubeconfig for the seed cluster. Ideally (for security reasons) this kubeconfig should not contain any credentials (e.g. only oidc or auth-provider configuration), but we do not make any assumptions as of now.
For compatibility reasons we will fallback to `<seed-name>.oidc`. We will remove the fallback in a future `gardenctl` release.

Fixes #

**Release note**:

```feature user

```

### DIFF
--- a/internal/gardenclient/client_test.go
+++ b/internal/gardenclient/client_test.go
@@ -1,0 +1,128 @@
+/*
+SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package gardenclient_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+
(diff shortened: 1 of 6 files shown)


internal/gardenclient/client_test.go
```diff
@@ -0,0 +1,127 @@
...
+		Namespace: "default",
+		AuthInfo:  "user",
+		Cluster:   "cluster",
+	}
+	config.CurrentContext = name
+	data, err := clientcmd.Write(*config)
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
+
+	return data
+}
```
petersutter, line 127: currently I have duplicated `createTestKubeconfig` from `target_suite_test.go`. Ideally we should move it to some test utils package. @holgerkoser I remember that you had a similar requirement in the past but couldn't find such a place in the code.

grolu (approved):
/lgtm

internal/gardenclient/client_test.go
petersutter: as discussed I have added a TODO

internal/gardenclient/client_test.go
```diff
@@ -0,0 +1,128 @@
...
+			})
+
+			It("it should fail with not found error", func() {
+				_, err := gardenClient.GetSeedClientConfig(ctx, seed)
+				Expect(err).To(HaveOccurred())
+				Expect(apierrors.IsNotFound(err)).To(BeTrue())
+			})
+		})
+	})
+})
```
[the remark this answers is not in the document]
holgerkoser: I would create a static set of seed secrets and access these data in different ways. For the assert function I would use a table test.

PR #100, opened 2022-02-11
Enh/target flag improvements

grolu:
This PR removes not working target flags override logic from root cmd command.
It also moves target flag registration from root to subcommands and introduces new target flags for the target command and subcommands.

Fixes #

**Release note**:

```feature user

```


internal/util/factory.go
```diff
@@ -35,6 +35,8 @@ type Factory interface {
 	// returned slice can contain IPv6, IPv4 or both, in no particular
 	// order.
 	PublicIPs(context.Context) ([]string, error)
+	// TF returns the target flags used to override the target
+	TF() target.TargetFlags
```
grolu, line 39: Not sure about extending the factory. Maybe there is a better way

pkg/cmd/base/options.go
```diff
@@ -58,18 +67,81 @@ func WrapRunE(o CommandOptions, f util.Factory) func(cmd *cobra.Command, args []
 	}
 }
 
+type CobraCompletionFunc func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective)
+type CobraCompletionFuncWithError func(ctx context.Context, manager target.Manager) ([]string, error)
+
+// WrapCompletionFunction creates a function that can be used to register a cobra completion command that has access to the manager
+func WrapCompletionFunction(f util.Factory, ioStreams util.IOStreams, completer CobraCompletionFuncWithError) CobraCompletionFunc {
```
grolu, line 74: I moved the completion wrapper here and exported it at it maybe can be reused by other completion functions that require the manager. Currently it is only used in the base options though

pkg/cmd/ssh/ssh_test.go
```diff
@@ -286,16 +287,15 @@ var _ = Describe("SSH Command", func() {
 			Expect(gardenClient.Get(ctx, key, bastion)).NotTo(Succeed())
 
 			// assert that no temporary SSH keypair remained on disk
-			_, err := os.Stat(options.SSHPublicKeyFile)
-			Expect(err).To(HaveOccurred())
+			//_, err := os.Stat(options.SSHPublicKeyFile)
```
grolu, line 290: need some help on this. The tests used values from the options that are set internally by the run command. We no longer have access as the options struct is not exposed.

pkg/cmd/ssh/ssh_test.go
```diff
@@ -348,18 +348,15 @@ var _ = Describe("SSH Command", func() {
...
```
grolu, line 351: same problem, see comment above

pkg/cmd/base/options.go
```diff
@@ -58,18 +67,81 @@ func WrapRunE(o CommandOptions, f util.Factory) func(cmd *cobra.Command, args []
...
 func NewOptions(ioStreams util.IOStreams) *Options {
 	return &Options{
 		IOStreams: ioStreams,
 	}
 }
 
-// AddFlags adds flags to adjust the output to a cobra command
-func (o *Options) AddFlags(flags *pflag.FlagSet) {
+// AddOutputFlags adds flags to adjust the output to a cobra command
+func (o *Options) AddOutputFlags(flags *pflag.FlagSet) {
```
holgerkoser, line 101:
```suggestion
// AddOutputFlag adds flag to adjust the output to a cobra command
func (o *Options) AddOutputFlag(flags *pflag.FlagSet) {
```

internal/util/factory.go
holgerkoser: Why dont't we use the same approach like kubectl does with https://pkg.go.dev/k8s.io/cli-runtime/pkg/genericclioptions.

pkg/cmd/base/options.go
holgerkoser: Why not a new package genericcompletions or at least a new file completions.

pkg/cmd/ssh/ssh_test.go
holgerkoser: For E2E tests create a cobra command and call `cmd.Execute()` and maybe use mocks or just check the output. In this case do not create a cobra command but create a `SSHOptions` instance and call the `Run` method.

pkg/cmd/base/options.go
holgerkoser: As I have already said. I think it would make sense to have first an agreed concept for this kind of PRs before doing all the implementation work.

pkg/cmd/base/options.go
```diff
@@ -58,18 +67,81 @@ func WrapRunE(o CommandOptions, f util.Factory) func(cmd *cobra.Command, args []
...
 	flags.StringVarP(&o.Output, "output", "o", o.Output, "One of 'yaml' or 'json'.")
 }
 
+// AddTargetOverrideFlags adds flags to override the current target to a cobra command
+func (o *Options) AddTargetOverrideFlags(f util.Factory, flags *pflag.FlagSet) {
```
holgerkoser, line 106: Why do we pass the factory here??

grolu:
/close
Close this PR as it is out of sync with master and changes need to be discussed. Keep branch for now, maybe we can use it when we continue work on this topic

PR #112, opened 2022-02-17
Fixes #28, #110 , #111, #113 and #114 

### LINKED ISSUE #28: Error running command
**What happened**:
An error was printed when running the completion command because the `gardenctl` configuration was not found

When running a command for which the gardenctl configuration is not required, I would not expect an error that the configuration is not found

Remove the `gardenctl` configuration and run the `completion` command

**Environment**:

### LINKED ISSUE #110: Config file not clear from error message
When running a command that requires a config file but the config is not present, `spf13/viper` returns the following error when trying to load the config
```bash
g target --garden dummy
E0217 16:19:01.020125   54088 cmd.go:175] failed to read config file: Config File "gardenctl-v2" Not Found in "[/Users/me/.garden]"
Error: failed to complete command options: failed to load config: failed to open file: open : no such file or directory
```
It states it could not find the file `gardenctl-v2` but in fact `viper` is looking for a file `gardenctl-v2.yaml`.

In case the config file is not found the error message should contain the correct filename / path so that it is clear for the user where to place it.

**Why is this needed**:
Config file not clear from error message

### LINKED ISSUE #111: GCTL_SESSION_ID should not be required for some commands
**What happened**:
```bash
gardenctl-v2 help
Error: Environment variable GCTL_SESSION_ID is required. Use "gardenctl help" for more information about the requirements of gardenctl
```

```bash
gardenctl-v2 rc bash
Error: Environment variable GCTL_SESSION_ID is required. Use "gardenctl help" for more information about the requirements of gardenctl
```

```bash
gardenctl-v2 version
Error: Environment variable GCTL_SESSION_ID is required. Use "gardenctl help" for more information about the requirements of gardenctl
```

Currently the above mentioned commands fail if there is no `GCTL_SESSION_ID` or `TERM_SESSION_ID`
Running these commands should succeed as they do necessarily require the `GCTL_SESSION_ID` to be present

**Environment**:

### LINKED ISSUE #113: completion not working in custom zsh setup
Hello,

Thanks for reaching out so quickly and helping in identifying the issue.
**Issue:**
When you run `gardenctl rc zsh` then it will generate the following script:
```bash
if (( $+commands[gardenctl] )); then
    gctl_completion_file="${fpath[1]}/_gardenctl"
    if [[ ! -f "$gctl_completion_file" ]]; then
        gardenctl completion zsh >| "$gctl_completion_file"
        source "$gctl_completion_file"
    else
        source "$gctl_completion_file"
        gardenctl completion zsh >| "$gctl_completion_file" &|
    fi
    unset gctl_completion_file
fi
```
If your $fpath contains a location (first one) which does not exist then this `gardenctl completion zsh` will never run. This was the first finding. I had temporarily changed this script to not take the path from $fpaths and the `gardenctl completion zsh` was run and sourced. However the completion still did not work.

It was found that cobra generates the completion script which has `compdef _gardenctl gardenctl` commented out. This cause the issue on my setup of zsh and it did not run the function `_gardenctl`. However when i uncommented this line in the script it started to work.

Please check if always enabling this would not be a problem in a vanilla zsh or other setups (whichever can be tested). If its safe to do it then you can make this change in the tool, else i will continue to use a function which will explicitly call this.

Best Regards,
Madhav

### LINKED ISSUE #114: Completion does not work for zsh and OMZ
When using [zsh](https://github.com/zsh-users/zsh) together with [OMZ](https://github.com/ohmyzsh/ohmyzsh) the completion does not work.
However when adding `compdef _gardenctl gardenctl` to `~/.zshrc` it solves the issue 

**Why is this needed**:

holgerkoser:
With this PR a not existing config file is treated like an empty config file. An initial config struct with a default filename is created if no config file exists. The user can create a new config file with:
```sh
gardenctl config set-garden dev --kubeconfig /path/to/kubeconfig
```
The initialization of the sessionDirectory and the targetFilename is done when the manager is created and not in the initConfig of the root command. The `version`, `help`, `completion` and `rc` do no more throw an error if `GCTL_SESSION_ID` is not set.

The help of the `gradenctl rc zsh` has been improved and the startup script has been adapted to also support zsh frameworks and plugin manager. Example configuration with gardenctl zsh custom plugin are described for [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh), [zgen](https://github.com/tarjoilija/zgen) and [zinit](https://github.com/zdharma-continuum/zinit) .

Fixes #28 
Fixes #110 
Fixes #111
Fixes #113
Fixes #114

**Release note**:

```feature user

```

### DIFF
--- a/docs/help/gardenctl_rc_bash.md
+++ b/docs/help/gardenctl_rc_bash.md
@@ -7,9 +7,11 @@ Generate a gardenctl startup script for bash
 Generate a gardenctl startup script for bash that contains various tweaks,
 such as setting environment variables, loading completions and adding some helpful aliases or functions.
 
-To load gardenctl startup script for each bash session, execute once:
+To load gardenctl startup script for each bash session add the following line at the end of the ~/.bashrc file:
 
-    echo 'source <(gardenctl rc bash)' >> ~/.bashrc
+    source <(gardenctl rc bash)
+
+You will need to start a new shell for this setup to take effect.
 
 
 ```
@@ -20,6 +22,7 @@ gardenctl rc bash [flags]
 
 ```
   -h, --help            help for bash
+      --no-completion   The startup script should not setup completion
   -p, --prefix string   The prefix used for aliases and functions (default "g")
 ```
 
--- a/docs/help/gardenctl_rc_fish.md
+++ b/docs/help/gardenctl_rc_fish.md
@@ -7,9 +7,11 @@ Generate a gardenctl startup script for fish
 Generate a gardenctl startup script for fish that contains various tweaks,
 such as setting environment variables, loading completions and adding some helpful aliases or functions.
 
-To load gardenctl startup script for each fish session, execute once:
+To load gardenctl startup script for each fish session add the following line at the end of the ~/.config/fish/config.fish file:
 
-    echo 'gardenctl rc fish | source' >> ~/.config/fish/config.fish
+    gardenctl rc fish | source
+
+You will need to start a new shell for this setup to take effect.
 
 
 ```
@@ -20,6 +22,7 @@ gardenctl rc fish [flags]
 
 ```
   -h, --help            help for fish
+      --no-completion   The startup script should not setup completion
   -p, --prefix string   The prefix used for aliases and functions (default "g")
 ```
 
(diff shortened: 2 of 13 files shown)


petersutter (approved):
/lgtm

PR #116, opened 2022-02-21
doc: Improve install from Github release

ialidzhikov:
There should be a reason why we have a version var (unused otherwise).

**Release note**:

```feature user
NONE
```


petersutter (approved):
/lgtm thanks

holgerkoser (approved):
/lgtm

PR #119, opened 2022-02-22
Ambiguous Shoot targeting should return an error

### LINKED ISSUE #117: No error on ambiguous targeting
**What happened**:
When just targeting a garden and a shoot (without specifying the project or the seed) it does not fail with an error if the same shoot name exists in another / multiple projects

When debugging, I could verify that `remainingItemCount` is not set
https://github.com/gardener/gardenctl-v2/blob/master/internal/gardenclient/client.go#L174-L177

**Environment**:

petersutter:
When targeting a shoot without specifying project or seed and there are multiple shoots on the `garden` with the same name an error should be returned.
This was not the case as [`remainingItemCount`](https://github.com/gardener/gardenctl-v2/blob/master/internal/gardenclient/client.go#L174-L177) was not set. From the documentation it says

> 	// [...] If the list request contained label or field selectors, then the number of
> 	// remaining items is unknown and the field will be left unset and omitted during serialization.

in our case the list request contained a field selector, hence `remainingItemCount` is unset.

Instead, we will fetch up to two items in order to figure out if there are multiple search results

Fixes #117 

We could also get rid of the `RemainingItemCount` handling as we anyhow increased the limit to `2`. Please let me know in case I should remove it

**Release note**:

```feature user

```

### DIFF
--- a/internal/gardenclient/client.go
+++ b/internal/gardenclient/client.go
@@ -160,7 +160,7 @@ func (g *clientImpl) GetShoot(ctx context.Context, namespace, name string) (*gar
 }
 
 func (g *clientImpl) FindShoot(ctx context.Context, opts ...client.ListOption) (*gardencorev1beta1.Shoot, error) {
-	opts = append(opts, client.Limit(1))
+	opts = append(opts, client.Limit(2))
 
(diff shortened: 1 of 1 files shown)


holgerkoser (approved):
/lgtm

PR #131, opened 2022-03-07
use referred shoot of managed seeds for targeting, provider-env and ssh command

### LINKED ISSUE #104: Enable `ssh` and `provider-env` when targeting Seed that is a ManagedSeed

When targeting a `Seed`, that is backed by a `ManagedSeed` (i.e. a `Shoot` registered as `Seed`), the `ssh` and `provider-env` commands cannot be used:
```
$ g target --garden dev seed aws
Successfully targeted seed "aws"

$ eval $(g kubectl-env zsh)

$ g target view
garden:"dev", seed:"aws"%

$ g ssh ip-10-11-12-13.eu-west-1.compute.internal
Error: no Shoot cluster targeted

$ eval $(g provider-env zsh)
Error: no shoot targeted
```

Obviously, for such `Seed`s you can switch the target to the respective `Shoot` and then use the `ssh` and `provider-env` commands.
However, it would be nice to support these commands out of the box, when targeting `ManagedSeed`s.

**Why is this needed**:

Typically, most `Seed`s in a gardener installation (at least the user-facing ones) are `ManagedSeed`s.
When operators start investigating some issues with a shoot control plane or seed, they would typically target the `Seed`. Later on, they might need to `ssh` into a node of the seed cluster or need to use the cloud-provider CLI for further investigation.

### LINKED ISSUE #129: Use
Currently, we require for all seeds to have secret with name `seedname.login` defined in the garden namespace.
However, for shooted seeds, we could simply get the shoot kubeconfig. We can look for `managedseed` resources in the garden namespace to find shoots that are used as seeds.

**Why is this needed**:
Reduces configuration overhead for shooted seeds.

grolu:
If seed is a managed seed, GetSeedClientConfig returns GetShootClientConfig for respective shoot

Fixes #129 Fixes #104

**Release note**:

```feature operator
Managed seeds no longer require `<seedname>.login` secret. The `gardenlogin` kubeconfig is used instead for the shoot that is registered as seed
```

```feature user
Gardenctl now uses referred shoot of managed seeds for targeting, provider-env and ssh commands.
```

### DIFF
--- a/internal/gardenclient/client.go
+++ b/internal/gardenclient/client.go
@@ -281,13 +281,7 @@ func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clie
 	if shoot, err := g.GetShootOfManagedSeed(ctx, name); err != nil {
 		return nil, err
 	} else if shoot != nil {
-		// TODO: Discuss if we want to have the fallback in case GetShootClientConfig returns an error or better just do:
-		// return g.GetShootClientConfig(ctx, "garden", shoot.Name)
-		if config, err := g.GetShootClientConfig(ctx, "garden", shoot.Name); err == nil {
-			return config, nil
-		} else if !apierrors.IsNotFound(err) {
-			return nil, err
-		}
+		return g.GetShootClientConfig(ctx, "garden", shoot.Name)
 	}
 
 	key := types.NamespacedName{Name: name}
--- a/internal/gardenclient/client.go
+++ b/internal/gardenclient/client.go
@@ -78,8 +78,8 @@ type Client interface {
 	// GetConfigMap returns a Kubernetes configmap resource
 	GetConfigMap(ctx context.Context, namespace, name string) (*corev1.ConfigMap, error)
 
-	// GetManagedSeed returns a Gardener Seed Management ManagedSeed resource
-	GetManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.ManagedSeed, error)
+	// GetShootOfManagedSeed returns shoot of seed using ManagedSeed resource, nil if not a managed seed
+	GetShootOfManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.Shoot, error)
 
 	// RuntimeClient returns the underlying kubernetes runtime client
 	// TODO: Remove this when we switched all APIs to the new gardenclient
@@ -262,31 +262,31 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
 	return cm, nil
 }
 
-func (g *clientImpl) GetManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.ManagedSeed, error) {
+func (g *clientImpl) GetShootOfManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.Shoot, error) {
 	managedSeed := &seedmanagementv1alpha1.ManagedSeed{}
-	key := types.NamespacedName{Name: name, Namespace: "garden"} // Currently, managed seeds are restricted to the garden namespace
+	key := types.NamespacedName{Namespace: "garden", Name: name} // Currently, managed seeds are restricted to the garden namespace
 
 	if err := g.c.Get(ctx, key, managedSeed); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+
 		return nil, fmt.Errorf("failed to get managed seed %v: %w", key, err)
 	}
 
-	return managedSeed, nil
+	return managedSeed.Spec.Shoot, nil
 }
 
 func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clientcmd.ClientConfig, error) {
-	managedSeed, err := g.GetManagedSeed(ctx, name)
-	if err != nil && !apierrors.IsNotFound(err) {
+	if shoot, err := g.GetShootOfManagedSeed(ctx, name); err != nil {
 		return nil, err
-	}
-
-	if managedSeed != nil {
-		clientConfig, err := g.GetShootClientConfig(ctx, managedSeed.Namespace, managedSeed.Spec.Shoot.Name)
-		if err != nil && !apierrors.IsNotFound(err) {
+	} else if shoot != nil {
+		// TODO: Discuss if we want to have the fallback in case GetShootClientConfig returns an error or better just do:
+		// return g.GetShootClientConfig(ctx, "garden", shoot.Name)
+		if config, err := g.GetShootClientConfig(ctx, "garden", shoot.Name); err == nil {
+			return config, nil
+		} else if !apierrors.IsNotFound(err) {
 			return nil, err
-		}
-
-		if clientConfig != nil {
-			return clientConfig, nil
 		}
 	}
 
--- a/internal/gardenclient/client.go
+++ b/internal/gardenclient/client.go
@@ -11,6 +11,8 @@ import (
 	"errors"
 	"fmt"
 
+	"k8s.io/klog/v2"
+
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -273,6 +275,8 @@ func (g *clientImpl) GetShootOfManagedSeed(ctx context.Context, name string) (*s
 
 		return nil, fmt.Errorf("failed to get managed seed %v: %w", key, err)
 	}
+
+	klog.V(1).Infof("using referred shoot %q for seed %q", managedSeed.Spec.Shoot.Name, name)
 
 	return managedSeed.Spec.Shoot, nil
 }
--- a/internal/gardenclient/client.go
+++ b/internal/gardenclient/client.go
@@ -11,11 +11,6 @@ import (
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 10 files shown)


internal/gardenclient/client.go
```diff
@@ -257,7 +258,27 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
 	return cm, nil
 }
 
+func (g *clientImpl) GetManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.ManagedSeed, error) {
+	managedSeed := &seedmanagementv1alpha1.ManagedSeed{}
+	key := types.NamespacedName{Name: name, Namespace: "garden"} // Currently, managed seeds are restricted to the garden namespace
+
+	if err := g.c.Get(ctx, key, managedSeed); err != nil {
+		return nil, fmt.Errorf("failed to get managed seed %v: %w", name, err)
```
petersutter, line 266:
```suggestion
		return nil, fmt.Errorf("failed to get managed seed %v: %w", key, err)
```

internal/gardenclient/client.go
```diff
@@ -257,7 +258,27 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
...
+}
+
 func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clientcmd.ClientConfig, error) {
+	managedSeed, err := g.GetManagedSeed(ctx, name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return nil, err
+	}
+
+	if managedSeed != nil {
+		return g.GetShootClientConfig(ctx, managedSeed.Namespace, managedSeed.Name)
```
petersutter, line 279: if it returns a notFound error (e.g. gardenlogin-controller-manager is not running) it should continue and fallback to the .login kubeconfig secret

pkg/cmd/env/options.go
```diff
@@ -129,15 +131,27 @@ func (o *options) Run(f util.Factory) error {
...
 
+		if t.ShootName() == "" {
+			if t.SeedName() != "" {
+				managedSeed, err := client.GetManagedSeed(f.Context(), t.SeedName())
+				if err != nil && !apierrors.IsNotFound(err) {
+					return err
+				}
+
+				if managedSeed != nil {
+					o.CurrentTarget = o.CurrentTarget.WithProjectName("garden").WithShootName(managedSeed.Name)
```
holgerkoser, line 148: The currentTarget should not be changed. Or we should rename the property to Target.

pkg/cmd/ssh/options.go
```diff
@@ -419,18 +421,29 @@ func (o *SSHOptions) Run(f util.Factory) error {
...
-
-	printTargetInformation(o.IOStreams.Out, currentTarget)
-
 	// create client for the garden cluster
 	gardenClient, err := manager.GardenClient(currentTarget.GardenName())
 	if err != nil {
 		return err
 	}
 
+	if currentTarget.ShootName() == "" {
```
holgerkoser, line 430: same here. The currentTarget can only be chnaged via the target command.

pkg/cmd/env/options.go
holgerkoser: The managedSeed has a namespace which we use in case of getting the clientConfig. Here we ignore the namespace an hardcode the the project name to be `garden`

grolu: Yes, I agree. I also think that we should rename it. If we leave the code at that place, we should definitively do it.

grolu: >The managedSeed has a namespace which we use in case of getting the clientConfig. Here we ignore the namespace an hardcode the the project name to be garden

The `Seed` resource is not namespaces, however the `ManagedSeed` resource is. Don't ask me why because it is only possible to create `ManagedSeed` resources in the garden namespaces. Of course, shoots that are used as seeds are (currently) always in the garden namespace.
See also here: https://github.com/gardener/gardener/blob/master/docs/usage/managed_seed.md
>Unlike the Seed resource, the ManagedSeed resource is namespaced. Currently, managed seeds are restricted to the garden namespace.

That's why I hard-coded it. Do you have a cleaner solution?

internal/gardenclient/client.go
holgerkoser: I think it is better to always return the same kind of kubeconfig for the same seed. It should be clear to user what he gets for a managed seed. If there is no `gardenlogin-controller-manager` we anyway have a problem.

internal/gardenclient/client.go
```diff
@@ -257,7 +262,34 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
...
+func (g *clientImpl) GetManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.ManagedSeed, error) {
+	managedSeed := &seedmanagementv1alpha1.ManagedSeed{}
+	key := types.NamespacedName{Name: name, Namespace: "garden"} // Currently, managed seeds are restricted to the garden namespace
+
+	if err := g.c.Get(ctx, key, managedSeed); err != nil {
+		return nil, fmt.Errorf("failed to get managed seed %v: %w", key, err)
+	}
+
+	return managedSeed, nil
+}
```
holgerkoser, line 274:
```suggestion
func (g *clientImpl) GetShootNameOfSeed(ctx context.Context, name string) (*types.NamespacedName, error) {
	managedSeed := &seedmanagementv1alpha1.ManagedSeed{}
	key := types.NamespacedName{
	      Namespace: "garden", // Currently, managed seeds are restricted to the garden namespace
	      Name: name,
	 }

	if err := g.c.Get(ctx, key, managedSeed); err != nil {
		if apierrors.IsNotFound(err) {
			return nil, nil
		}

		return nil, fmt.Errorf("failed to get managed seed %v: %w", key, err)
	}

	return &types.NamespacedName{
            Namespace: "garden", 
            Name: managedSeed.Spec.Shoot.Name,
          }, nil
}
```

internal/gardenclient/client.go
```diff
@@ -257,7 +262,34 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
...
+	if managedSeed != nil {
+		clientConfig, err := g.GetShootClientConfig(ctx, managedSeed.Namespace, managedSeed.Name)
+		if err != nil && !apierrors.IsNotFound(err) {
+			return nil, err
+		}
+
+		if clientConfig != nil {
+			return clientConfig, nil
+		}
+	}
```
holgerkoser, line 291:
```suggestion
	if shootName, err := g.GetShootNameofSeed(ctx, name); err != nil {
		return nil, err
	} else if shootName != nil {
	        // I would prefer the following
	        // return g.GetShootClientConfig(ctx, "garden", *shootName)
	        clientConfig, err := g.GetShootClientConfig(ctx, "garden", *shootName)
	        if err == nil {
                       return clientConfig, nil
                } else if !apierrors.IsNotFound(err) {
                       return nil, err
                }
	}
```

internal/gardenclient/client_test.go
```diff
@@ -64,6 +119,8 @@ var _ = Describe("Client", func() {
 		},
 			Entry("and the secretName has suffix .oidc", "seed-1"),
 			Entry("and the secretName has suffix .login", "seed-2"),
+			Entry("and seed is a managed seed, return shoot client config", "managedSeed-1"),
```
holgerkoser, line 122: This should not be added to this table test. If we have a managed seed the secret doesn't matter. may be for this case anyway the table test make not much sense. I would prefer to use a helper with ExpectOffset and create or override the resources for each test as needed.

internal/gardenclient/client.go
grolu: I changed it as @petersutter requested (as he had good arguments) but I kind of also agree with @holgerkoser.
Let's discuss this again.

pkg/cmd/env/options.go
grolu: I renamed the current target

pkg/cmd/ssh/options.go
grolu: I renamed the current target

internal/gardenclient/client.go
grolu: I incorporated  the changes. We still need to discuss the fallback to seed kubeconfig in case of an error.

internal/gardenclient/client.go
grolu: removed fallback as discussed

pkg/cmd/ssh/options.go
```diff
@@ -413,29 +414,37 @@ func (o *SSHOptions) Run(f util.Factory) error {
...
+		if sshTarget.SeedName() != "" {
+			if shoot, err := gardenClient.GetShootOfManagedSeed(f.Context(), sshTarget.SeedName()); err != nil {
+				return err
+			} else if shoot != nil {
+				sshTarget = sshTarget.WithProjectName("garden").WithShootName(shoot.Name)
+			} else {
+				return target.ErrNoShootTargeted
+			}
+		}
+	}
```
petersutter, line 439: `if sshTarget.ShootName() == "" && sshTarget.SeedName() != "" {`
```suggestion
	if sshTarget.ShootName() == "" && sshTarget.SeedName() != "" {
		if shoot, err := gardenClient.GetShootOfManagedSeed(f.Context(), sshTarget.SeedName()); err != nil {
			return err
		} else if shoot != nil {
			sshTarget = sshTarget.WithProjectName("garden").WithShootName(shoot.Name)
		} else {
			return target.ErrNoShootTargeted
		}
	}
```

grolu:
closed in favor of #156

PR #144, opened 2022-03-29
Initialize KUBECONFIG environment variable in startup script

holgerkoser:
At the end of the startup script the KUBECONFIG environment variable is set to `kubeconfig.yaml` symlink in the current session directory. This behavior can be disabled by using the `--no-kubeconfig` flag. 

Fixes #

**Release note**:

```feature user
Initialize KUBECONFIG environment variable in startup script using `rc` command. You can disable this bahavior by using the `--no-kubeconfig` flag
```

### DIFF
--- a/pkg/cmd/env/templates/rc.tmpl
+++ b/pkg/cmd/env/templates/rc.tmpl
@@ -3,16 +3,16 @@ if [ -z "$GCTL_SESSION_ID" ] && [ -z "$TERM_SESSION_ID" ]; then
   export GCTL_SESSION_ID=$(uuidgen)
 fi
 alias {{.prefix}}=gardenctl
-{{if not .noCompletion -}}
-source <(gardenctl completion {{.shell}})
-complete -o default -F __start_gardenctl {{.prefix}}
-{{end -}}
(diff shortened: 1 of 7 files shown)


pkg/cmd/env/templates/rc.tmpl
```diff
@@ -2,9 +2,9 @@
 if [ -z "$GCTL_SESSION_ID" ] && [ -z "$TERM_SESSION_ID" ]; then
   export GCTL_SESSION_ID=$(uuidgen)
 fi
+alias {{.prefix}}=gardenctl
```
petersutter, line 5: we should have a consistent order across the shells
e.g. on zsh the alias is defined after the {{if not .noCompletion -}} block

holgerkoser: The completions work is very different in all shell. To have the same order in all shells I will move the completions behind the aliases. It would be strange to define completion for an alias that not has been defined yet even if it would work.

holgerkoser: ✔️

petersutter (approved):
/lgtm

grolu (approved):
/lgtm

PR #147, opened 2022-04-01
SSH to node that did not join the cluster

### LINKED ISSUE #146: Can't SSH to node that has not yet joined the cluster
**What happened**:
There is a validation that aborts the SSH command in case the node was not found: https://github.com/gardener/gardenctl-v2/blob/master/pkg/cmd/ssh/options.go#L466-L471

The operation should not abort in case the node did not yet join the cluster as that could also be the reason why the user wants to troubleshoot this node.

An information could be printed in case the node was not found but it should still continue.

**Environment**:

petersutter:
gardenctl continues with the SSH operation in case the node was not found (e.g. node failed to joined the cluster)

Fixes #146

**Release note**:

```bugfix user
Fixed an issue where it was not possible to SSH to a node that has not yet joined the cluster
```

### DIFF
--- a/pkg/cmd/ssh/options.go
+++ b/pkg/cmd/ssh/options.go
@@ -40,6 +40,7 @@ import (
 	"golang.org/x/crypto/ssh/agent"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/cmd/ssh/options.go
```diff
@@ -463,11 +464,17 @@ func (o *SSHOptions) Run(f util.Factory) error {
...
+	if o.NodeName != "" {
+		nodeHostname, err = getNodeHostname(ctx, o, shootClient, o.NodeName)
+		if err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to determine hostname for node: %w", err)
+			}
+
+			fmt.Fprintf(o.IOStreams.Out, "Node with name %q not found. Wrong name provided or this node did not yet join the cluster, continuing anyways: %v\n", o.NodeName, err)
+			nodeHostname = o.NodeName
+		}
```
holgerkoser, line 477:
```suggestion
	var nodeHostname string
	if o.NodeName != "" {
          node, err := getShootNode(ctx, o, shootClient, currentTarget)
          if err == nil {
            nodeHostname, err = getNodeHostname(node)
            if err != nil {
              return err
            }
          } else if apierrors.IsNotFound(err) {
            fmt.Fprintf(o.IOStreams.Out, "Node with name %q not found. Wrong name provided or this node did not yet join the cluster, continuing anyways: %v\n", o.NodeName, err)
            nodeHostname = o.NodeName
          } else {
            return fmt.Errorf("failed to determine hostname for node: %w", err)
          }
```

I think it would be more clear to not mix `getNodeHostname` and `getShootNode`

holgerkoser: I not sure if this proposal is correct but it should give you the idea.

pkg/cmd/ssh/options.go
```diff
@@ -1036,7 +1025,12 @@ func writeToTemporaryFile(key []byte) (string, error) {
 	return f.Name(), nil
 }
 
-func getNodeHostname(node *corev1.Node) (string, error) {
+func getNodeHostname(ctx context.Context, o *SSHOptions, shootClient client.Client, nodeName string) (string, error) {
```
holgerkoser, line 1028: I would keep the old signature

pkg/cmd/ssh/options.go
petersutter: ✅

holgerkoser (approved):
/lgtm

PR #148, opened 2022-04-13
Get seed client: fail on any other error than not found

petersutter:
This PR fixes a nil pointer dereference when targeting a seed cluster. This happens e.g. in case you are not authorized to read the `<seed>.login` secret in the `garden` namespace.

Fixes #

**Release note**:

```bugfix user
Fixes a nil pointer dereference when targeting a seed cluster. This happens e.g. in case you are not authorized to read the `<seed>.login` secret in the `garden` namespace.
```


grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #149, opened 2022-04-13
Generate Shoot gardenlogin kubeconfig locally

petersutter:
**What this PR does / why we need it**:
With https://github.com/gardener/gardener/pull/5612 we now have the ability to generate the Shoot `gardenlogin` kubeconfig locally, which this PR is about.

It does not rely anymore on the `<shoot>.kubeconfig` ConfigMap that is managed by the `gardenlogin-controller-manager`. This means we can deprecate the `gardenlogin-controller-manager` component.

**Which issue(s) this PR fixes**:
Fixes #128

**Special notes for your reviewer**:
To keep this PR "small" it does not contain the command to print out the Shoot kubeconfig. This should be done in a separate PR.

**Release note**:

```feature user
Shoot cluster kubeconfigs are now generated locally instead of relying on the `<shoot>.kubeconfig` ConfigMap that was created by the `gardenlogin-controller-manager`.
```

PR #150, opened 2022-04-14
Print target kubeconfig

petersutter:
This PR adds the command `target kubeconfig` to print the kubeconfig for the current target.

This is especially useful to retrieve the gardenlogin kubeconfig for `Shoot` clusters, as these kubeconfigs will not be written anymore to the garden cluster as ConfigMaps, as the `gardenlogin-controller-manager` will be deprecated (ref #149)

Fixes #

**Release note**:

```feature user
You can now print the kubeconfig for the current target using `gardenctl target kubeconfig` command
```

### DIFF
--- a/docs/help/gardenctl_target_kubeconfig.md
+++ b/docs/help/gardenctl_target_kubeconfig.md
@@ -1,28 +1,15 @@
-## gardenctl target
+## gardenctl target kubeconfig
 
-Set scope for next operations, using subcommands or pattern
-
-```
-gardenctl target [flags]
-```
-
-### Examples
+Print the kubeconfig for the current target
 
 ```
-# target project "my-project" of garden "my-garden"
-gardenctl target --garden my-garden --project my-project
-
-# target shoot "my-shoot" of currently selected project
-gardenctl target shoot my-shoot
-
-# Target shoot control-plane using values that match a pattern defined for a specific garden
-gardenctl target value/that/matches/pattern --control-plane
+gardenctl target kubeconfig [flags]
 ```
 
 ### Options
 
 ```
-  -h, --help            help for target
+  -h, --help            help for kubeconfig
   -o, --output string   One of 'yaml' or 'json'.
 ```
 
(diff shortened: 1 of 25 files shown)


pkg/cmd/target/kubeconfig.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return err
+	}
+
+	b, err := writeRawConfig(config)
+	if err != nil {
+		return err
+	}
+
+	_, err = o.IOStreams.Out.Write(b)
```
holgerkoser, line 89: fmt.Fprintln(o.IOStreams.Out.Write, string(b))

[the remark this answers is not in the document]
holgerkoser: or ignore the error. If this no works the user does no see errors
```go
fmt.Fprintf(o.IOStreams.Out, "%s\n", b)

return nil
```

pkg/cmd/target/target.go
```diff
@@ -48,6 +48,7 @@ gardenctl target value/that/matches/pattern --control-plane`,
 
 	cmd.AddCommand(NewCmdUnset(f, NewUnsetOptions(ioStreams)))
 	cmd.AddCommand(NewCmdView(f, NewViewOptions(ioStreams)))
+	cmd.AddCommand(NewCmdKubeconfig(f, NewKubeconfigOptions(ioStreams)))
```
holgerkoser, line 51: I think it is better to have this command at top level. It is a command the uses the current target and does something with this information like `ssh` or `k-env` or `p-env`. They are also not subcommands of target.

holgerkoser: Or we create a new top level `get` command (like kubectl) and add this one as first subcommand. In the future we could add a `target` subcommand and deprecate `target view`.

docs/help/gardenctl_target_kubeconfig.md
```diff
@@ -0,0 +1,43 @@
...
+
+```
+gardenctl target kubeconfig [flags]
+```
+
+### Options
+
+```
+  -h, --help            help for kubeconfig
+  -o, --output string   One of 'yaml' or 'json'.
```
holgerkoser, line 13: Handle the output flag in run

pkg/cmd/target/kubeconfig.go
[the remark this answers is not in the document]
petersutter: ✅

pkg/cmd/target/target.go
petersutter: ✅

PR #156, opened 2022-05-03
use referred shoot of managed seeds for targeting, provider-env and ssh command #2

### LINKED ISSUE #104: Enable `ssh` and `provider-env` when targeting Seed that is a ManagedSeed

### LINKED ISSUE #129: Use

grolu:
If seed is a managed seed, GetSeedClientConfig returns GetShootClientConfig for respective shoot

Fixes #129 Fixes #104

**Release note**:

```feature operator
Managed seeds no longer require `<seedname>.login` secret. The `gardenlogin` kubeconfig is used instead for the shoot that is registered as seed
```

```feature user
Gardenctl now uses referred shoot of managed seeds for targeting, provider-env and ssh commands.
```

### DIFF
--- a/pkg/cmd/ssh/options.go
+++ b/pkg/cmd/ssh/options.go
@@ -26,14 +26,6 @@ import (
 	"sync"
 	"time"
 
-	gardencorev1alpha1 "github.com/gardener/gardener/pkg/apis/core/v1alpha1"
-	corev1alpha1helper "github.com/gardener/gardener/pkg/apis/core/v1alpha1/helper"
-	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
-	corev1beta1constants "github.com/gardener/gardener/pkg/apis/core/v1beta1/constants"
-	operationsv1alpha1 "github.com/gardener/gardener/pkg/apis/operations/v1alpha1"
-	"github.com/gardener/gardener/pkg/utils"
-	gutil "github.com/gardener/gardener/pkg/utils/gardener"
-	"github.com/gardener/gardener/pkg/utils/secrets"
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh"
 	cryptossh "golang.org/x/crypto/ssh"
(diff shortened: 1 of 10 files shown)


pkg/cmd/ssh/options.go
```diff
@@ -416,35 +417,43 @@ func (o *SSHOptions) Run(f util.Factory) error {
...
 	// create client for the garden cluster
-	gardenClient, err := manager.GardenClient(currentTarget.GardenName())
+	gardenClient, err := manager.GardenClient(sshTarget.GardenName())
 	if err != nil {
 		return err
 	}
 
+	if sshTarget.ShootName() == "" && sshTarget.SeedName() != "" {
+		if shoot, err := gardenClient.GetShootOfManagedSeed(f.Context(), sshTarget.SeedName()); err != nil {
+			return err
```
petersutter, line 434: if isNotFoundError -> return error "cannot ssh to non-managed Seeds"
else -> return err

something like that

grolu: like this?

holgerkoser: wrap error

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #163, opened 2022-09-30
Doc/gardenctl config

etiennnr:
There was missing documentation on how to configure gardenctl if the kube-system namespace is blocked for a user.

So added a 2nd method to do that over CLI. I get and know that this is not an ideal solution, but it works well in the meantime

Fixes #

**Release note**:

```doc user
Added a second method to configure gardenctl which requres less premissions
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ export KUBECONFIG=~/relative/path/to/kubeconfig.yaml
 # Method 1 : Fetch cluster-identity of garden cluster from the configmap
 export CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
 # OR
-# Method 2 : If you don't have access to the kube-system namespace in the garden, the garden cluster-identity is also available in every shoot's yaml
+# Method 2 : If you don't have access to the kube-system namespace in the garden cluster, the garden cluster-identity can also be extracted from every shoot's yaml
 export PROJECT="your-project-name" # Change to your project name
 export SHOOT="your-shoot-name" # Change to any shoot's name in your project
 export PREFIX="shoot--$PROJECT--$SHOOT-"$(kubectl get shoot -n garden-$PROJECT $SHOOT -ojsonpath={.metadata.uid})"-"
--- a/README.md
+++ b/README.md
@@ -59,13 +59,19 @@ You can modify this file directly using the `gardenctl config` command. It allow
 Example `config` command:
 ``` bash
 # Adapt the path to your kubeconfig file for the garden cluster
-export KUBECONFIG=~/relative/path/to/kubeconfig.yaml
+❯ export KUBECONFIG=~/relative/path/to/kubeconfig.yaml
 
-# Fetch cluster-identity of garden cluster
-CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# Method 1 : Fetch cluster-identity of garden cluster from the configmap
+❯ export CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# OR
+# Method 2 : If you don't have access to the kube-system namespace in the garden, the garden cluster-identity is also available in every shoot's 
+❯ export SHOOT_UID=$(kubectl get shoot -n garden-my-project my-shoot -ojsonpath={.metadata.uid})
+❯ export PREFIX="shoot--my-project--my-shoot-$SHOOT_UID-"
+❯ kubectl -n garden-my-project get shoot my-shoot -ojsonpath={.status.clusterIdentity}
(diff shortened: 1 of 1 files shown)


etiennnr:
@petersutter do you think it would also be a good idea to somehow add the landscape name (or garden cluster identity) in the dashboard?

petersutter:
> @petersutter do you think it would also be a good idea to somehow add the landscape name (aka garden cluster identity) in the dashboard?

yes, I have created https://github.com/gardener/dashboard/issues/1314

README.md
```diff
@@ -61,8 +61,15 @@ Example `config` command:
 # Adapt the path to your kubeconfig file for the garden cluster
 export KUBECONFIG=~/relative/path/to/kubeconfig.yaml
 
-# Fetch cluster-identity of garden cluster
-CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# Method 1 : Fetch cluster-identity of garden cluster from the configmap
+export CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# OR
+# Method 2 : If you don't have access to the kube-system namespace in the garden, the garden cluster-identity is also available in every shoot's yaml
```
petersutter, line 67:
```suggestion
# Method 2 : If you don't have access to the kube-system namespace in the garden cluster, the garden cluster-identity can also be extracted from every shoot's yaml
```

README.md
```diff
@@ -61,8 +61,15 @@ Example `config` command:
 # Adapt the path to your kubeconfig file for the garden cluster
 export KUBECONFIG=~/relative/path/to/kubeconfig.yaml
 
-# Fetch cluster-identity of garden cluster
-CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# Method 1 : Fetch cluster-identity of garden cluster from the configmap
+export CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
```
petersutter, line 65: it's not necessary to `export` variables, except for `KUBECONFIG`

etiennnr: Good point!

etiennnr:
@petersutter @grolu Please also merge as I don't have the correct rights :)

README.md
```diff
@@ -61,8 +61,15 @@ Example `config` command:
...
-# Fetch cluster-identity of garden cluster
+# Method 1 : Fetch cluster-identity of garden cluster from the configmap
 CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonpath={.data.cluster-identity})
+# OR
+# Method 2 : If you don't have access to the kube-system namespace in the garden cluster, the garden cluster-identity can also be extracted from every shoot's yaml
+PROJECT="your-project-name" # Change to your project name
+SHOOT="your-shoot-name" # Change to any shoot's name in your project
+PREFIX="shoot--$PROJECT--$SHOOT-"$(kubectl get shoot -n garden-$PROJECT $SHOOT -ojsonpath={.metadata.uid})"-"
+STATUS=$(kubectl get shoot -n garden-$PROJECT $SHOOT -ojsonpath={.status.clusterIdentity})
+CLUSTER_IDENTITY=$(echo ${STATUS#$PREFIX}) # difference between both
```
holgerkoser, line 72:
```suggestion
# This function should be added to your shell profile
function getClusterIdentity () {
  local ns=$(kubectl get project $1 -ojsonpath={.spec.namespace});
  local uid=$(kubectl get shoot -n $ns $2 -ojsonpath={.metadata.uid}); 
  local status=$(kubectl get shoot -n $ns $2 -ojsonpath={.status.clusterIdentity});
  local prefix="shoot--$1--$2-$uid-"; 
  echo ${status#"$prefix"}; 
}
CLUSTER_IDENTITY=$(getClusterIdentity "your-project-name" "your-shoot-name")
```

etiennnr: Why adding a function in the shell profile? This will only be needed once as it's just to configure gardenctl...

After that, it's going to be saved in `~/.garden/gardenctl-v2.yaml`

petersutter: I also don't think that we need the function in the shell profile, this should only be a simple example on how to configure / setup gardenctl.

If we really want to extract it to a function it should handle both cases (as you mentioned out of this thread) and first try to fetch it from the kube-system namespace and if it fails ask the user to input a project and shoot name to try the fallback. Only then the project name and shoot name is relevant.

But I'm not sure if it's worth the effort / makes it more complicated

holgerkoser: Ok. But we should not have something in our documentation where the namespace is constructed by prepending `garden-` to the project name. This must be changed.

etiennnr: That's indeed a fair point! Just added an updated commit

petersutter (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #175, opened 2022-11-02
Add cidr flag completion

### LINKED ISSUE #67: Add `cidr` flag completion on `ssh` command
`cidr` flag completion for the `ssh` command.

The completion should return:
- the public `IP`s using the ipify service
- the `IP`s of the systems network interfaces, excluding `down` and `loopback` interfaces.

This adds some convenience for the user to specify the correct `CIDR` range that should be allowed to access the bastion. Alternatively the user should still be able to specify a different `CIDR` that is not part of the completion (e.g. because of natting).

**Why is this needed**:
When running the `ssh` command you can pass the `cidr` range as flag that should be allowed to access the bastion via flag. For more details on the `SSH`-flow see [GEP-15](https://github.com/gardener/gardener/blob/master/docs/proposals/15-manage-bastions-and-ssh-key-pair-rotation.md#ssh-flow)

If it is not provided, `gardenctl` tries to guess the proper `IP`/`CIDR` range by using a public service (https://www.ipify.org/ in particular).
However using the public IP may not work in all scenarios, e.g. when using a VPN to access non-internet facing bastions. In this case, the user should be able to specify the correct `CIDR` range that should be used and `gardenctl` should have a completion for the `cidr` flag that would add some convenience.

Out of scope: 
- The IP could be automatically detected similar to the public `ipify` service. Instead, such a service would have to be setup within the internal network and `gardenctl` would call it to have it return its internal `IP`s.
- First create the bastion to get it's "public" endpoint from the `status`. Then figure out which network interface is used by using something like `ip route get <bastion-endpoint>`. Afterwards alter the `bastion` resource to specify the `IP` that was returned. See also https://engineering.qubecinema.com/2019/05/13/go-routing-package.html which is using `github.com/google/gopacket`, however not all architectures are supported.

petersutter:
This PR adds `cidr` flag completion for the `ssh` command.

The completion returns:
- the public `IP`s as cidr using the ipify service
- the cidrs of the systems network interfaces, excluding `down` and `loopback` interfaces.

This adds some convenience for the user to specify the correct `CIDR` range that should be allowed to access the bastion. Alternatively the user should still be able to specify a different `CIDR` that is not part of the completion (e.g. because of natting).

Fixes #67

**Release note**:

```feature user
`--cidr` flag completion now available for the `ssh` command
```

### DIFF
--- a/pkg/cmd/ssh/ssh.go
+++ b/pkg/cmd/ssh/ssh.go
@@ -8,8 +8,10 @@ package ssh
 
 import (
 	"fmt"
+	"net"
 
 	"github.com/spf13/cobra"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 
 	"github.com/gardener/gardenctl-v2/internal/util"
 	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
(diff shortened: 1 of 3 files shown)


pkg/cmd/ssh/ssh.go
```diff
@@ -43,5 +45,74 @@ func NewCmdSSH(f util.Factory, o *SSHOptions) *cobra.Command {
...
+	includeFlags := net.FlagUp
+	excludeFlags := net.FlagLoopback
+
+	for _, iface := range interfaces {
+		addrs, err := iface.Addrs()
+		if err != nil {
+			return nil, err
+		}
+
+		if is(iface, includeFlags) && isNot(iface, excludeFlags) {
```
holgerkoser, line 101: What is happening here? Why this bit operation?

petersutter: I want to trim down the list of suggestions and only include interfaces that are `up` and exlude loopback interfaces.
Or what do you want to know?
See also https://cs.opensource.google/go/go/+/master:src/net/interface.go;l=36-73?q=FlagBroadcast&ss=go%2Fgo

holgerkoser (approved):
/lgtm

PR #180, opened 2022-11-07
enh: add cmd to patch bastion ipblock/cidr filter

### LINKED ISSUE #123: Update own (public) IP of Bastion (SSH)
There should be a way that `gardenctl` updates the field `spec.ingress[].ipBlock` of the `Bastion` resource in case the own IP has changed (e.g. now connected via VPN) without having to create a new `Bastion`

**Why is this needed**:

sven-petersen:

Introduce a new command that allows the update, currently only the CIDR/IPBlock, of an existing bastion host previously created through the `ssh` command.

Fixes #123 

* Name of the command needs to be decided on
* The new command currently is part of the ssh package as they are closely related. Maybe makes sense to put it in its own package?
* Probably the way suggestions for --bastion-name are created, by trying to determine the users name and list bastions created by that user only, needs to be tested further

**Release note**:

```feature user
Introduce new command, `ssh-patch`, to update existing bastion resources
```


pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,343 @@
...
+		}
+
+		return "", nil
+	}
+)
+
+//nolint:revive
+type SSHPatchOptions struct {
+	sshBaseOptions
+	// the patch target
```
petersutter, line 57:
```suggestion
	// BastionName is the patch target
```
doc comments for types start with naming the declared symbol

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,343 @@
...
+func (o *SSHPatchOptions) Validate() error {
+	if err := o.sshBaseOptions.Validate(); err != nil {
+		return err
+	}
+
+	if o.BastionName == "" {
+		return fmt.Errorf("bastion name flag not provided and unable to automatically determine bastion name")
+	}
+
+	return nil
```
petersutter, line 342: shouldn't it be also validated that there are cidrs set on the options? Otherwise this command wouldn't do anything

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,343 @@
...
+	hasChanges, _, err := patchBastionResource(o, bastion)
+
+	if err != nil {
+		return err
+	} else if !hasChanges {
+		return nil
+	}
+
+	return gardenClient.PatchBastion(ctx, bastion, oldBastion)
+}
```
petersutter, line 232: in addition to my other [comment](https://github.com/gardener/gardenctl-v2/pull/180/files#r1015745823), I would inline most of the extracted functions, as it adds imo unneeded complexity
I have also proposed to change the function name to `patchBastionIngressPolicies`
```go
func patchBastionIngressPolicies(ctx context.Context, o *SSHPatchOptions, gardenClient gardenClient.Client, sshTarget target.Target) error {
	bastion, err := getBastion(ctx, o, gardenClient, sshTarget) // TODO bastion should be taken from the options struct
	if err != nil {
		return err
	}

	oldBastion := bastion.DeepCopy()

	var policies []operationsv1alpha1.BastionIngressPolicy

	for _, cidr := range o.CIDRs {
		if *bastion.Spec.ProviderType == "gcp" {
			ip, _, err := net.ParseCIDR(cidr)
			if err != nil {
				return err // this should never happen, as it is already checked within the Validate function
			}

			if ip.To4() == nil {
				if !o.AutoDetected {
					return fmt.Errorf("GCP only supports IPv4: %s", cidr)
				}

				fmt.Fprintf(o.IOStreams.Out, "GCP only supports IPv4, skipped CIDR: %s\n", cidr)

				continue // skip
			}
		}

		policies = append(policies, operationsv1alpha1.BastionIngressPolicy{
			IPBlock: networkingv1.IPBlock{
				CIDR: cidr,
			},
		})
	}

	if len(policies) == 0 {
		return errors.New("there are no ingress policies to be applied")
	}

	bastion.Spec.Ingress = policies

	return gardenClient.PatchBastion(ctx, bastion, oldBastion)
}
```

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,343 @@
...
+	gardenClient, err := manager.GardenClient(sshTarget.GardenName())
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	return manager, sshTarget, gardenClient, nil
+}
+
+func (o *SSHPatchOptions) Run(f util.Factory) error {
+	_, sshTarget, gardenClient, err := getGardenClient(f)
```
petersutter, line 255: In the long run do not want to pass the factory anymore to the `Run` function, hence for new commands we shouldn't use it anymore.
Instead, put everything you need for the `Run` function on the options struct.

pkg/cmd/ssh/ssh_patch_test.go
```diff
@@ -0,0 +1,386 @@
...
...ampleClientCertficate = []byte("-----BEGIN CERTIFICATE-----\nMIIDITCCAgmgAwIBAgIIVteacLIfMJkwDQYJKoZIhvcNAQELBQAwFTETMBEGA1UE\nAxMKa3ViZXJuZXRlczAeFw0yMjExMDQwNzQ1MjVaFw0yMzExMDQwNzQ1MjZaMDQx\nFzAVBgNVBAoTDnN5c3RlbTptYXN0ZXJzMRkwFwYDVQQDExBrdWJlcm5ldGVzLWFk\nbWluMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwbmPW7NqesUYGtIB\nXOfGrmMeXsWxkM0/sT0LXOZ6c7N/EnRrFAUKMrpMJd0jpePzBN96aLNl08UXpKRj\nHjawqoVEZxcRM6m2sGSeRXCHSz3j1ptkvk2H+fAbJybZ4YvQtT0da1IXxWeBOEvM\nimefH4821KL0MPvOyZLWQaAMZPjvPxqI9nbpOK84KPJ5xGJzOO1lFMC6bNi5QCWG\nypyXWyIjZTsyThu3zNP484NUlE/eixalixGJNKNmMVnzaaytEp/fr0QEzA10HNFV\nemOpsU2r4Yp94GfHpZsmpnUSlUIyWsgFXnJkQNSfHiP4b+5nojwhxTS5KGCiI0cZ\n+ZS82QIDAQABo1YwVDAOBgNVHQ8BAf8EBAMCBaAwEwYDVR0lBAwwCgYIKwYBBQUH\nAwIwDAYDVR0TAQH/BAIwADAfBgNVHSMEGDAWgBQd3FiThv+dW40xiwfqbv9bK6Tl\nrzANBgkqhkiG9w0BAQsFAAOCAQEACAVjFMu7htdP9EoHVzZeJIx4SukwDbBeMxIx\nu5hZfQRBDdVTwaias+RiUHeFsyknbcLXvmYRTncr48FJtQozfoYvmFI7AVMBRalO\n7zmhOOsGPKnDDaUWzbo05d4zprMg8x7PWpPJTrnD+ebcNpA+H6K77wg76PaX/vb4\nHUpop7TEo/jsBaaI3AZoJDYXwmij0zjQFUc9E7XZ/p+V345vAIezxolaMHlJZbYv\naolz2dNFv2zTJg1CWkgozWNTY/BaTaJzO4oKVxLRHZWHTrXAU/MRlaRguZQ1tW0f\nS48mSaUhmlEw3l+GE+xjEWe+1pYjZEl2mTNUcOOSMe+hI07E3A==\n-----END CERTIFICATE-----\n")
```
petersutter, line 47: I'd rather not hardcode the client certificate, instead I would generate it like so https://github.com/gardener/gardenlogin/blob/c3e7c7cb38fec2854e3d0564d2e14c3ffcde3448/cmd/get_client_certificate_test.go#L480-L493 and serialize it if needed.
This looks like a real client certificate, but probably from a local kind cluster.
If you hardocde one, make sure that the CN etc. clearly indicate that this is a test certificate. Leave other fields empty if not really needed, like the organization, but at least it shouldn't have the special `system:masters` group.

petersutter (commented):
part 2

pkg/cmd/ssh/ssh_patch_options.go
sven-petersen: @petersutter The CIDRs and the validation logic was moved to the "base options". So CIDRs should be checked through `o.sshBaseOptions.Validate()`. Is that approach ok for you?

petersutter: ✅

pkg/cmd/ssh/ssh_patch.go
```diff
@@ -0,0 +1,42 @@
...
+			if err != nil {
+				fmt.Fprintln(o.IOStreams.ErrOut, err.Error())
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			}
+
+			return bastionNames, cobra.ShellCompDirectiveNoFileComp
+		},
+	}
+
+	cmd.Flags().StringArrayVar(&o.CIDRs, "cidr", o.CIDRs, "CIDRs to allow access to the bastion host; if not given, your system's public IPs (v4 and v6) are auto-detected.")
```
holgerkoser, line 39: Implement this in AddFlags on options
```go
func (o *SSHPatchOptions) AddFlags(flags *pflag.FlagSet) {
	flags.StringArrayVar(&o.CIDRs, "cidr", o.CIDRs, "CIDRs to allow access to the bastion host; if not given, your system's public IPs (v4 and v6) are auto-detected.")
}
```

pkg/cmd/ssh/ssh_patch_export_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	gc "github.com/gardener/gardenctl-v2/internal/gardenclient"
+	"github.com/gardener/gardenctl-v2/internal/util"
+	"github.com/gardener/gardenctl-v2/pkg/target"
+)
+
+var getCurrentUserOriginal = getCurrentUser
+
+func SetGetCurrentUser(f func(ctx context.Context, gardenClient gc.Client, authInfo *api.AuthInfo) (string, error)) {
```
holgerkoser, line 23: Use gomock to mock methods

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,400 @@
...
+)
+
+//nolint:revive
+type SSHPatchOptions struct {
+	sshBaseOptions
+
+	// BastionName is the patch targets name
+	BastionName string
+	// Factory gives access to the gardenclient and target manager
+	Factory util.Factory
```
holgerkoser, line 113: I think we should not store the factory on options. We have to fix the Run method later.

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,400 @@
...
+	if len(policies) == 0 {
+		return errors.New("no ingress policies could be created")
+	}
+
+	o.Bastion.Spec.Ingress = policies
+
+	return gardenClient.PatchBastion(ctx, o.Bastion, oldBastion)
+}
+
+func getManagerTargetAndGardenClient(f util.Factory) (target.Manager, target.Target, gardenClient.Client, error) {
```
holgerkoser, line 243: This is a strange reuse function. I would inline the code

pkg/cmd/ssh/ssh_patch_options.go
petersutter: I suggested:

> [petersutter](https://github.com/petersutter) [7 days ago](https://github.com/gardener/gardenctl-v2/pull/180#discussion_r1016693643)
> In the long run do not want to pass the factory anymore to the Run function, hence for new commands we shouldn't use it anymore.
> Instead, put everything you need for the Run function on the options struct.

https://github.com/gardener/gardenctl-v2/pull/180#discussion_r1016693643
but I didn't mean the factory itself :) maybe this was a bit misleading

pkg/cmd/ssh/ssh_patch_options.go
petersutter: +1
I hoped this one would vanish after putting the needed properties on the options in the complete function (inlined)

pkg/cmd/ssh/ssh_patch_export_test.go
```diff
@@ -0,0 +1,55 @@
...
+	return getAuthInfo
+}
+
+func GetGetBastionNameCompletions(o *SSHPatchOptions) func(f util.Factory, cmd *cobra.Command, toComplete string) ([]string, error) {
+	return o.getBastionNameCompletions
+}
+
+var timeNowOriginal = timeNow
+
+func SetTimeNow(f func() time.Time) {
```
holgerkoser, line 45: We have a clock interface on the factory

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	if bastion.ObjectMeta.UID == "" {
+		return nil, fmt.Errorf("Bastion '%s' in namespace '%s' not found", o.BastionName, shoot.Namespace)
+	}
+
+	return bastion, nil
+}
+
+func (o *sshPatchOptions) getAuthInfo(ctx context.Context) (*api.AuthInfo, error) {
```
petersutter, line 75:
```suggestion
func (o *sshPatchOptions) getAuthInfo() (*api.AuthInfo, error) {
```

pkg/cmd/ssh/ssh_patch_options.go
```diff
@@ -176,15 +91,15 @@ func getAuthInfo(ctx context.Context, manager target.Manager) (*api.AuthInfo, er
...
 	var bastionOfUser []*gardenoperationsv1alpha1.Bastion
 
-	currentUser, err := getCurrentUser(ctx, gardenClient, authInfo)
+	currentUser, err := o.Utils.GetCurrentUser(ctx, o.gardenClient, o.authInfo)
 	if err != nil {
 		return nil, err
 	}
 
-	list, err := gardenClient.ListBastions(ctx)
+	list, err := o.gardenClient.ListBastions(ctx)
```
petersutter, line 102:
```
g ssh-patch
Error: failed to complete command options: failed to list bastions with list options []: bastions.operations.gardener.cloud is forbidden: User "system:serviceaccount:garden-core:myrobot" cannot list resource "bastions" in API group "operations.gardener.cloud" at the cluster scope
```
listing bastions across all namespaces will fail for non-operators

pkg/cmd/ssh/ssh_patch_utils.go
```diff
@@ -0,0 +1,166 @@
...
+func (u *sshPatchUtilsImpl) GetBastionsOfUser(ctx context.Context, user string, gardenClient gardenClient.Client, bastionListOption client.ListOption) ([]*gardenoperationsv1alpha1.Bastion, error) {
+	var bastionsOfUser []*gardenoperationsv1alpha1.Bastion
+
+	list, err := gardenClient.ListBastions(ctx, bastionListOption)
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range list.Items {
+		bastion := list.Items[i]
```
petersutter, line 140:
```suggestion
	for _, bastion := range list.Items {
```

pkg/cmd/ssh/base_options.go
```diff
@@ -0,0 +1,72 @@
...
+	CIDRs []string
+
+	// AutoDetected indicates if the public IPs of the user were automatically detected.
+	// AutoDetected is false in case the CIDRs were provided via flags.
+	AutoDetected bool
+}
+
+func (o *sshBaseOptions) Complete(f util.Factory, cmd *cobra.Command, args []string) error {
+	if len(o.CIDRs) == 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
```
holgerkoser, line 33:
```suggestion
		ctx, cancel := context.WithTimeout(f.Context(), 60*time.Second)
```

pkg/cmd/sshpatch/export_test.go
```diff
@@ -0,0 +1,61 @@
...
+			target:       target,
+			clientConfig: clientConfig,
+		},
+	}
+}
+
+type TestOptions struct {
+	options
+	Out     *util.SafeBytesBuffer
+	Streams util.IOStreams
```
holgerkoser, line 39: Why do we store the streams agai. We already have IOStreams?

pkg/cmd/sshpatch/completions.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/gardener/gardenctl-v2/internal/util"
+)
+
+type completions struct{}
+
+func newCompletions() *completions {
```
holgerkoser, line 16: Do we need this constructor?

pkg/cmd/sshpatch/completions.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/gardener/gardenctl-v2/internal/util"
+)
+
+type completions struct{}
```
holgerkoser, line 14: Why do we need this struct?

pkg/cmd/ssh/base_options.go
```diff
@@ -0,0 +1,72 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/gardener/gardenctl-v2/internal/util"
+	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
+)
+
+// BaseOptions is a struct used by all ssh related commands
+type BaseOptions struct {
```
holgerkoser, line 18: I would rename this this to `AllowAddressRanges` or something like this

pkg/cmd/sshpatch/export_test.go
```diff
@@ -0,0 +1,61 @@
...
+				},
+			},
+		},
+		Out:     out,
+		Streams: streams,
+	}
+}
+
+//nolint:revive
+func NewTestCompletions() *completions {
```
holgerkoser, line 59: Not needed!!

pkg/cmd/sshpatch/export_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/gardener/gardenctl-v2/pkg/cmd/base"
+	"github.com/gardener/gardenctl-v2/pkg/cmd/ssh"
+	"github.com/gardener/gardenctl-v2/pkg/target"
+)
+
+type TestUserBastionListPatcherImpl struct {
+	userBastionListPatcherImpl
+}
+
+func NewTestUserBastionPatchLister(manager target.Manager) *TestUserBastionListPatcherImpl {
```
holgerkoser, line 22: Should not be in export_test

pkg/cmd/sshpatch/ssh_patch.go
```diff
@@ -0,0 +1,44 @@
+package sshpatch
```
holgerkoser, line 1: rename this file to `sshpatch`

pkg/cmd/sshpatch/ssh_patch_test.go
```diff
@@ -0,0 +1,446 @@
+/*
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package sshpatch_test
```
holgerkoser, line 7: rename this file to `sshpatch_test`

holgerkoser: There is not `sshpatch_suite_test` in this package. Please create this file.

pkg/cmd/sshpatch/export_test.go
holgerkoser: Move it to the `sshpatch_suite_test` instead

sven-petersen: Should `NewTestOptions` then moved as well?

holgerkoser (approved):
/lgtm

PR #185, opened 2022-11-22
Enh/target cluster by alias

### LINKED ISSUE #121: Name (alias) for garden configuration
Currently, when targeting a garden without using the patterns feature, you have to specify the garden cluster identity, which could be a lengthy name or even a guid. It should be possible, to specify a name for the garden that can be used as alternative when targeting the garden cluster. E.g. the configuration could look like

```yaml
gardens:
- identity: my-landscape-dev
  name: dev
  kubeconfig: ~/relative/path/to/kubeconfig.yaml
```

You should then be able to run `gardenctl target --garden dev` and of course also `gardenctl target --garden my-landscape-dev`

**Why is this needed**:
Users should be able to name and target the garden to their convenience, especially when the garden cluster identity is meaningless (e.g. a GUID) or rather lengthy.

sven-petersen:

Fixes #121 

**Release note**:

```feature user
Allow targeting of clusters by using an alias instead of the identity
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -99,7 +99,44 @@ func LoadFromFile(filename string) (*Config, error) {
 		config.LinkKubeconfig = &val
 	}
 
+	if err := config.validate(); err != nil {
+		return nil, err
+	}
+
 	return config, nil
+}
+
+// validate checks the config for static errors. It does not check if the kubeconfig file exists.
+func (config *Config) validate() error {
+	seen := make(map[string]bool, len(config.Gardens))
+
+	for i := range config.Gardens {
+		garden := config.Gardens[i]
+
+		if garden.Name == "" {
+			return fmt.Errorf("identity is required for gardens in the gardenctl configuration")
+		}
+
+		if garden.Kubeconfig == "" {
+			return fmt.Errorf("kubeconfig is required for gardens in the gardenctl configuration")
+		}
+
+		if seen[garden.Name] {
+			return fmt.Errorf("identity and alias must be unique but %q was found multiple times in gardenctl configuration", garden.Name)
+		}
+
+		if garden.Alias != "" {
+			if seen[garden.Alias] {
+				return fmt.Errorf("identity and alias must be unique but %q was found multiple times in gardenctl configuration", garden.Alias)
+			}
+
+			seen[garden.Alias] = true
+		}
+
+		seen[garden.Name] = true
+	}
+
+	return nil
 }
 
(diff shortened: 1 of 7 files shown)


gardener-robot-ci-3:

pkg/config/config.go
```diff
@@ -96,9 +99,46 @@ func LoadFromFile(filename string) (*Config, error) {
 		config.LinkKubeconfig = &val
 	}
 
+	if err := config.validate(); err != nil {
+		return nil, err
+	}
```
petersutter, line 104: this will break gardenctl if there is a faulty garden config, e.g. one entry does not contain a kubeconfig. This should not be the case. 

In the example below you are not even able to correct the "mistake"
```
g config set-garden foo
Successfully configured garden "foo"
```

```
g config set-garden foo --kubeconfig /foo/bar
Error: failed to complete command options: failed to get target manager: failed to load config: kubeconfig is required for gardens in the gardenctl configuration
```

Similar to the `kubectl config` command, we don't want to be so strict. Maybe we even want to move it to a separate `config validate` command? Let's discuss this

pkg/config/config.go
```diff
@@ -148,14 +188,23 @@ func (config *Config) GardenNames() []string {
...
+// Garden returns the first matching Garden cluster by name (identity or alias) from the list
+// of configured Gardens
 func (config *Config) Garden(name string) (*Garden, error) {
-	i, ok := config.IndexOfGarden(name)
-	if !ok {
+	var gardenConfig *Garden
+
+	for idx := range config.Gardens {
+		cfg := &config.Gardens[idx]
+		if name == cfg.Name || name == cfg.Alias {
```
petersutter, line 198: not sure where to put this comment:

```
gardenctl target --garden dev --project myproject --shoot mycluster
Successfully targeted shoot "mycluster"
```

```bash
cat $KUBECONFIG

apiVersion: v1
clusters:
- cluster:
    extensions:
    - extension:
        gardenClusterIdentity: dev
        shootRef:
          name: mycluster
          namespace: garden-myproject
      name: client.authentication.k8s.io/exec
    server: https://api.example.com
# ...
```

the `gardenClusterIdentity` contains the alias, not the identity. Either we update `gardenlogin` to also support the alias, or we should only set the identity here

sven-petersen: I will have a look into this. Imho it would be better/cleaner to always use the identity here.

petersutter (approved):
/lgtm (but requires rebase)

holgerkoser (approved):
/lgtm

Issue #195, opened 2022-12-07
target: only target flags which higher specificity shall be available for subcommands


sven-petersen:
**What would you like to be added**:

Currently the target flags `--garden`, `--project`, `--shoot`, `--seed`, `--control-plane` are available for the `target` and all of its subcommands.

Clarify if

1) we want to keep it this way and allow sth. like `target project myproject --garden mygarden` OR
2) the target flags should only be available if none of the target-subcommand is used (`target --garden mygarden`)
**3) only target flags with a higher specificity are allowed. So `target project myproject --shoot myshoot` would be ok but not `target project myproject --garden mygarden` would not.**

**Why is this needed**:

follow up for [#85](https://github.com/gardener/gardenctl-v2/issues/85)

petersutter:
Only target flags with lower specificity should be allowed (1.)

PR #196, opened 2022-12-08
Refactor/global target flags

### LINKED ISSUE #85: Introduce subcommands for gardenctl target
Currently, `gardenctl target` has no real subcommands for the different targeting types.
This comes with several drawbacks:
- Subcommands not listed in help
- No dedicated help for different subcommands
- complex code needs to differentiate between target types

After introducing the subcommands, we can also make the global target flags lokal, adding them only to the commands where they make sense. In case of target command, also differentiate between subcommands, as some of the target flags may not make sense for a specific target type. E.g. something like `gardenctl target shoot foo --shoot bar` should not be supported.

sven-petersen:

Instead of registering the target flags (`--garden`, `--shoot`, ...) on the root cmd they are now registered per command. This way the flags are only available where they have an effect.

Fixes #85 (2nd step)

follow up: https://github.com/gardener/gardenctl-v2/issues/195 

**Release note**:

```other developer
Register target-flags only on cmds that require them instead of globally
```


gardener-robot-ci-1:

pkg/target/manager.go
```diff
@@ -31,10 +32,28 @@ var (
...
 )
 
+// AutoCompletionsFuncs provides a set of functions that lists suitable values for
+// targeting while taking the currently targeted garden account.
+// In case the manager gets refactored/split up these functions should stay together.
+type AutoCompletionsFuncs interface {
+	// ShootNames returns all shoots for the current target.
+	ShootNames(ctx context.Context) ([]string, error)
+	// SeedNames returns all seeds for the current target.
+	// The current target must at least point to a garden.
```
petersutter, line 42: > `// The current target must at least point to a garden.`

This applies all functions except `GardenNames`. Either we remove it here (as we hinted it already in the interface description or add it also for the other functions

pkg/target/manager_test.go
```diff
@@ -612,4 +612,107 @@ var _ = Describe("Target Manager", func() {
 			})
 		})
 	})
+
+	Context("AutoCompletionsFuncs", func() {
+		// var clientConfig clientcmd.ClientConfig
```
petersutter, line 617:
```suggestion
```

petersutter (approved):
/lgtm (but requires rebase)

pkg/flags/target.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+
+	"github.com/gardener/gardenctl-v2/internal/util"
+	"github.com/gardener/gardenctl-v2/pkg/target"
+)
+
+func RegisterTargetFlagCompletionFuncs(cmd *cobra.Command, factory util.Factory, ioStreams util.IOStreams, flags *pflag.FlagSet) {
```
holgerkoser, line 21: Rename to `RegisterTargetFlagsCompletionFuncs` or RegisterCompletionFuncsForTargetFlags`

holgerkoser (approved):
/lgtm
Do forget to rebase

PR #206, opened 2023-01-04
enh: target cmd - only register target flags with higher specificty

### LINKED ISSUE #195: target: only target flags which higher specificity shall be available for subcommands

sven-petersen:

see linked issue

Fixes #195 

**Release note**:

```breaking user
Only allow target flags (`--garden`, ...) with lower specificity when using `target` command
```


gardener-robot-ci-2:

petersutter (commented):
nit

pkg/cmd/target/seed.go
```diff
@@ -36,5 +38,11 @@ gardenctl target seed my-seed --garden my-garden`,
 
 	o.AddFlags(cmd.Flags())
 
+	manager, err := f.Manager()
+	utilruntime.Must(err)
+	manager.TargetFlags().AddGardenFlag(cmd.Flags())
+	manager.TargetFlags().AddProjectFlag(cmd.Flags())
```
petersutter, line 44:
```suggestion
```
remove

pkg/cmd/target/view.go
```diff
@@ -35,6 +37,11 @@ func NewCmdView(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
 
 	o.AddFlags(cmd.Flags())
 
+	manager, err := f.Manager()
+	utilruntime.Must(err)
+	manager.TargetFlags().AddFlags(cmd.Flags())
```
petersutter, line 42: not sure if we need target flags on the view command

petersutter (approved):
/lgtm

pkg/cmd/target/control-plane.go
```diff
@@ -36,5 +38,13 @@ gardenctl target control-plane --garden my-garden --project my-project --shoot m
 
 	o.AddFlags(cmd.Flags())
 
+	manager, err := f.Manager()
+	utilruntime.Must(err)
```
holgerkoser, line 42: It is not acceptable to assert no erros here. We never have done this anywhere else.

internal/util/factory.go
```diff
@@ -63,13 +66,19 @@ type FactoryImpl struct {
...
-	// TargetFlags can be used to completely override the target configuration
+	// targetFlags can be used to completely override the target configuration
 	// stored on the filesystem via a CLI flags.
-	TargetFlags target.TargetFlags
+	targetFlags target.TargetFlags
 }
 
 var _ Factory = &FactoryImpl{}
 
+func NewFactory() *FactoryImpl {
```
holgerkoser, line 76: Since this returns the strcut it should be named `NewFactoryImpl`?

pkg/flags/target_test.go
```diff
@@ -263,10 +260,7 @@ var _ = Describe("Target flags", func() {
 		var factory *util.FactoryImpl
 
 		BeforeEach(func() {
-			factory = &util.FactoryImpl{
-				TargetFlags: targetFlags,
-				ConfigFile:  configFile,
-			}
+			factory = util.NewFactory()
```
holgerkoser, line 263:
```
factory = util.NewFactory()
factory.ConfigFile = configFile
```

internal/util/factory.go
```diff
@@ -63,13 +66,19 @@ type FactoryImpl struct {
...
+	return &FactoryImpl{
+		targetFlags: target.NewTargetFlags("", "", "", "", false),
+	}
+}
```
holgerkoser, line 80:
```go
func NewFactoryImpl(tf target.TargetFlags) *FactoryImpl {
        var targetFlags target.TargetFlags
        if (tf != nil) {
           targetFlags = tf
        } else {
           targteFlags = target.NewTargetFlags("", "", "", "", false)
        }
	return &FactoryImpl{
		targetFlags,
	}
}
```

pkg/flags/target_test.go
holgerkoser: We agreed, that we don't need it and for new tests we can use the mockFactory.

holgerkoser (approved):
/lgtm

PR #246, opened 2023-03-03
`ssh`: Add flag to skip availability check for bastion

petersutter:
`ssh` subcommand: You can now skip the availability check for the bastion host using the flag `--skip-availability-check`. This is useful when you want to spin up a bastion host but `gardenctl` can't reach the bastion and thus would not be able to check for the availability.

Fixes #

**Release note**:

```feature user
`ssh` subcommand: You can now skip the availability check for the bastion host using the flag `--skip-availability-check`. This is useful when you want to spin up a bastion host but `gardenctl` can't reach the bastion and thus would not be able to check for the availability.
```


tedteng:
I observed the Gardenctl does not stick to the bastion instance yet a long time ago.    for example, 
1) g ssh nodeA --keep-bastion 
2) exit the current bastion
3) g ssh nodeA again, it will create an extra new bastion and ssh instance instead of sticky the previous bastion instance.  

Is this case already fixed? if yes please ignore the following information 

Is the purpose of this flag is accelerate ssh access as bastion has already been created, if yes,  I'm not quite sure the flag whether works as expected here as `g ssh nodeA --skip-availability-check` a new bastion is still being created and waiting for the instance build-up, but the flag will skip the check. Shall I create a new ticket about the stick bastion instance issue as in the case mentioned above or this flag will handle it?

petersutter:
> Is the purpose of this flag is accelerate ssh access as bastion has already been created

No, this was not the purpose, but it might come in handy in case we support to re-use the bastion instead of re-creating it.
We have a case where the bastion is not directly reachable from the users machine, so the availability check would fail and the bastion creation would time out after `10m`. This flag gives the user means to skip the availability check.

holgerkoser (approved):
/lgtm

sven-petersen (approved):
lgtm

PR #249, opened 2023-03-10
`ssh`: Add flag to disable keepalive

petersutter:
This PR adds a `--no-keepalive` flag that controls if the command should exit after the bastion becomes available.
If this option is true, no SSH connection will be established and the bastion will not be kept alive after it became available.
This option can only be used if KeepBastion is set to true and Interactive is set to false.

Fixes #

**Release note**:

```feature user
`ssh`: You can now disable the keepalive using `--no-keepalive` flag. The command exits after the bastion host became available without keeping the bastion alive or establishing an SSH connection. Note that this flag requires the flags `--interactive=false` and `--keep-bastion` to be set
```

### DIFF
--- a/pkg/cmd/ssh/options.go
+++ b/pkg/cmd/ssh/options.go
@@ -291,6 +291,12 @@ type SSHOptions struct {
 	// SkipAvailabilityCheck determines whether to check for the availability of
 	// the bastion host.
 	SkipAvailabilityCheck bool
+
+	// NoKeepalive controls if the command should exit after the bastion becomes available.
+	// If this option is true, no SSH connection will be established and the bastion will
+	// not be kept alive after it became available.
+	// This option can only be used if KeepBastion is set to true and Interactive is set to false.
+	NoKeepalive bool
 }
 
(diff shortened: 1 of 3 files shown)


grolu (approved):
/lgtm

pkg/cmd/ssh/options.go
```diff
@@ -151,6 +151,104 @@ var (
 
 		return cmd.Run()
 	}
+
+	// waitForSignal informs the user about their SSHOptions and keeps the
+	// bastion alive until gardenctl exits.
+	waitForSignal = func(ctx context.Context, o *SSHOptions, shootClient client.Client, bastion *operationsv1alpha1.Bastion, nodeHostname string, nodePrivateKeyFiles []string, signalChan <-chan struct{}) error {
```
holgerkoser, line 157: As far as I understand the code this function makes no assumption that the role of the cluster is a shoot cluster. I would rename the parameter from `shootClient` to `client` and remove the word shoot in all messages. What do you think?

holgerkoser (approved):
/lgtm

PR #257, opened 2023-03-15
`ssh`: Abort keepalive on not found

### LINKED ISSUE #232: Abort keepalive bastion if not found
**What happened**:
When running the `g ssh` command it creates a bastion resource and continuously sets the keepalive annotation on the bastion.
However if the bastion was deleted in the meantime (e.g. the local machine was hibernated) it prints out
```
Failed to keep bastion alive: bastions.operations.gardener.cloud "cli-zj4efox3" not found
Failed to keep bastion alive: bastions.operations.gardener.cloud "cli-zj4efox3" not found
...
```

In case of not found error, it should abort with a corresponding message instead of trying to annotate the bastion over and over again that does not exist anymore.

**Environment**:

petersutter:
The bastion keepalive is now aborted in case the bastion was deleted

Fixes #232

**Release note**:

```bugfix user
`ssh`: the command will now exit in case the `Bastion` resource was deleted
```

### DIFF
--- a/pkg/cmd/ssh/options.go
+++ b/pkg/cmd/ssh/options.go
@@ -544,7 +544,7 @@ func (o *SSHOptions) Run(f util.Factory) error {
 	}
 
 	// continuously keep the bastion alive by renewing its annotation
-	go keepBastionAlive(ctx, gardenClient.RuntimeClient(), bastion.DeepCopy())
+	go keepBastionAlive(ctx, cancel, gardenClient.RuntimeClient(), bastion.DeepCopy())
 
 	logger.Info("Waiting for bastion to be ready…", "waitTimeout", o.WaitTimeout)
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/ssh/options.go
```diff
@@ -151,6 +151,103 @@ var (
 
 		return cmd.Run()
 	}
+	// waitForSignal informs the user about their SSHOptions and keeps the
+	// bastion alive until gardenctl exits.
+	waitForSignal = func(ctx context.Context, o *SSHOptions, shootClient client.Client, bastion *operationsv1alpha1.Bastion, nodeHostname string, nodePrivateKeyFiles []string, signalChan <-chan struct{}) error {
```
holgerkoser, line 156: Seeee my comment in the other ssh PR.

holgerkoser (approved):
/lgtm

pkg/cmd/ssh/options.go
petersutter: as just discussed we leave it as is for now until we get feedback from users that it is confusing when targeting managed seed (for which there is always a shoot) and running ssh

PR #279, opened 2023-04-26
Refactor ssh command builder

petersutter:
- This PR refactors the code responsible for constructing the SSH command arguments
- Deduplicated code: Now, there is a single implementation for building the ssh command arguments, which is used for both printing to stdout and passing to `exec.CommandContext`
- The `ssh` command that is printed to stdout is now shell-escaped

Fixes #

**Release note**:

```other user
`ssh`: the command that is printed to stdout is now shell-escaped
```


holgerkoser (approved):
/lgtm

PR #283, opened 2023-04-27
`ssh` add `--bastion-name` to reuse bastion

petersutter:
add `--bastion-name` flag. If a bastion with this name doesn't exist, it will be created. If it does exist, the provided public SSH key must match the one used during the bastion's creation.
In addition the `--private-key-file` flag was added. It must be provided alongside the `--public-key-file` flag if you want to use a custom keypair. If not provided, gardenctl will either generate a temporary keypair or rely on the user's SSH agent for an available private key.

Fixes #

**Release note**:

```feature user
`ssh`: reuse bastion, e.g. `gardenctl ssh --keep-bastion --bastion-name cli-xxxxxxxx --public-key-file /path/to/ssh/key.pub --private-key-file /path/to/ssh/key`
- `--bastion-name` flag was added. If a bastion with this name doesn't exist, it will be created. If it does exist, the provided public SSH key must match the one used during the bastion's creation
- `--private-key-file` flag was added. It must be provided alongside the `--public-key-file` flag if you want to use a custom keypair. If not provided, gardenctl will either generate a temporary keypair or rely on the user's SSH agent for an available private key.
```

### DIFF
--- a/docs/help/gardenctl_ssh.md
+++ b/docs/help/gardenctl_ssh.md
@@ -31,11 +31,15 @@ gardenctl ssh
 # Create the bastion and output the connection information in JSON format
 gardenctl ssh --no-keepalive --keep-bastion --interactive=false --output json
 
+# Reuse a previously created bastion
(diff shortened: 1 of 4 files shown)


grolu (approved):
/lgtm

PR #284, opened 2023-05-02
`ssh`: add `--bastion-host`, `--bastion-port` flags to override the defaults

petersutter:
This PR introduces two new flags, `--bastion-host` and `--bastion-port`, to the `ssh` subcommand. These flags enable user to customize the bastion host and port, for example, in SSH port forwarding scenarios.

Fixes #

**Release note**:

```feature user
`ssh`: Use the `--bastion-host` and `--bastion-port` flags to customize the bastion host and port for the SSH client command, respectively. These flags are useful when you need to specify an alternative host and port for SSH port forwarding scenarios.
- `--bastion-host`: Specify a custom hostname or IP address for the bastion used in the SSH client command. If not provided, the address will be automatically determined.
- `--bastion-port`: Set the SSH port of the bastion used for the SSH client command. By default, this value is set to port 22.
```

### DIFF
--- a/docs/help/gardenctl_ssh.md
+++ b/docs/help/gardenctl_ssh.md
@@ -39,7 +39,9 @@ gardenctl ssh --keep-bastion --bastion-name cli-xxxxxxxx --public-key-file /path
 ### Options
 
 ```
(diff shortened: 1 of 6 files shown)


holgerkoser (approved):
/lgtm

Issue #225, opened 2023-01-30
ssh InsecureIgnoreHostKey seems not working 


tedteng:
**What happened**:
scenario：

This occurs in a limited IP address range more frequently. (eg CCEE for our case).
a new bastion created for the first time with IP `10.A.A.243`, an ssh know_hosts generated `10.A.A.243  ssh-ed25519 host key` persistent in local `/.ssh/known_hosts` file after `gardenctl ssh` successfully.
Then bastion destroy.  After some time, if a new bastion creates and the same IP `10.A.A.243` is assigned. it will fail for the ssh process as host key is different at this time, and the user can not access the cluster node. 
The `ssh.InsecureIgnoreHostKey()` has already been implemented.  however, gardenctl ssh process seems not working.  
https://github.com/gardener/gardenctl-v2/blob/faab11b02daebd75acf2265910e283aa16f5ae6b/pkg/cmd/ssh/options.go#L111-L116

![image](https://user-images.githubusercontent.com/42234376/215402535-e12f3530-f3d1-4262-9fc0-4b213ea20e1a.png)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:
1 ) `--keep-bastion` create a bastion 
2) g ssh node 
3) change local `~/.ssh/known_hosts` VM fingerprint to wrong one
3 ) g ssh node again
**Anything else we need to know**:

**Environment**:

/assign

tedteng:
will have a look next week when I vocation back.

tedteng:
creating fake VM fingerprints to check this issue, the issue was magically dismissed. I doubt it might be related to the nested network from CCEE before.  
![image](https://user-images.githubusercontent.com/42234376/221732669-5b9fd863-6bf2-4738-8440-3103886e2fcd.png)
close the issue.

PR #285, opened 2023-05-03
`ssh`: add `--bastion-user-known-hosts-file` flag

petersutter:
This PR adds a `--bastion-user-known-hosts-file` flag to the `ssh` command, allowing users to specify a custom known hosts file for the SSH connection to the bastion host.
This is useful when the bastion host IP is reused, which can lead to a failed remote host key verification if there is an existing entry for the same IP but from a previous (old) bastion.

Fixes #

Related issue #225

**Release note**:

```feature user
`ssh`: Use the `--bastion-user-known-hosts-file` flag to specify a custom known hosts file for the SSH connection to the bastion host. This is useful when the bastion host IP is reused, which can lead to a failed remote host key verification if there is an existing entry for the same IP but from a previous (old) bastion.
```


holgerkoser (approved):

holgerkoser (approved):
/lgtm

grolu (approved):
/lgtm

PR #289, opened 2023-05-05
Add target flags to `target` `view` command

### LINKED ISSUE #286: Target view command should also consider target flags
Target view command should also consider target flags, similar like it is handled in other commands.

**Why is this needed**:

petersutter:
This PR adds target flags to the `target` `view` command. 

This is beneficial when gardenctl is used within a script, and the script needs to determine the current target while also taking the target flags into account.

Fixes #286

**Release note**:

```feature user
The `target` `view` command (alias `gtv`) now also considers the target flags
```


pkg/cmd/target/view.go
```diff
@@ -35,6 +36,9 @@ func NewCmdView(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
 
 	o.AddFlags(cmd.Flags())
 
+	f.TargetFlags().AddFlags(cmd.Flags())
+	flags.RegisterCompletionFuncsForTargetFlags(cmd, f, ioStreams, cmd.Flags())
```
holgerkoser, line 40: Not related to this PR but why do we pass the flagset to this function. It is not used. The flags must already be registered. And I think it was a very bad idea to name this package `flags` which always create a name conflict if you want to have variable named flags. I anyway think it should be part of the same package the TargetFlags belongs to.

holgerkoser (approved):
/lgtm

pkg/cmd/target/view.go
petersutter: true, I have created an issue for it https://github.com/gardener/gardenctl-v2/issues/296

PR #298, opened 2023-05-11
`provider-env`: support `json` / `yaml` output

### LINKED ISSUE #287: `provider-env` command should support json / yaml output
Add --output flag to the `provider-env` command.

`provider-env` currently outputs a shell script when used like 

```bash
gardenctl provider-env zsh
```

**Why is this needed**:
For usage in scripts it would be beneficial to access the data in json or yaml format.

petersutter:
This PR adds the `--output` flag to the `provider-env` command

Fixes #287

**Release note**:

```feature user
`provider-env`: You can now control the output format using `--output` flag. The data that is typically passed to the templating engine for generating provider-env scripts is instead output in the specified format. Usage: `gardenctl provider-env -oyaml`
```

### DIFF
--- a/pkg/cmd/providerenv/options.go
+++ b/pkg/cmd/providerenv/options.go
@@ -162,10 +162,10 @@ func (o *options) Run(f util.Factory) error {
 		return err
 	}
 
-	return outputProviderEnv(o, shoot, secret, cloudProfile, messages)
+	return printProviderEnv(o, shoot, secret, cloudProfile, messages)
 }
 
-func outputProviderEnv(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret, cloudProfile *gardencorev1beta1.CloudProfile, messages ac.AccessRestrictionMessages) error {
+func printProviderEnv(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret, cloudProfile *gardencorev1beta1.CloudProfile, messages ac.AccessRestrictionMessages) error {
 	providerType := shoot.Spec.Provider.Type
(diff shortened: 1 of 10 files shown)


pkg/cmd/providerenv/options.go
```diff
@@ -153,10 +162,10 @@ func (o *options) Run(f util.Factory) error {
 		return err
 	}
 
-	return execTmpl(o, shoot, secret, cloudProfile, messages)
+	return outputProviderEnv(o, shoot, secret, cloudProfile, messages)
 }
 
-func execTmpl(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret, cloudProfile *gardencorev1beta1.CloudProfile, messages ac.AccessRestrictionMessages) error {
+func outputProviderEnv(o *options, shoot *gardencorev1beta1.Shoot, secret *corev1.Secret, cloudProfile *gardencorev1beta1.CloudProfile, messages ac.AccessRestrictionMessages) error {
```
holgerkoser, line 168: I would use `print` instead of `output` as verb for this function. The term output could be both the verb and the objects whch are printed. To me it more refers to the object(s) which are printed. Also in kubectl they more often use always `print` I think.
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L458-L559
https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/get/get.go#L260-L262

petersutter: ✅

holgerkoser (approved):
/lgtm

Issue #299, opened 2023-05-12
introduce plug-in feature in gardenctlv2  


tedteng:
**What would you like to be added**:
inspired by ticket https://github.com/gardener/gardenctl-v2/issues/12 introduce plug-in feature in gardenctlv2  
**Why is this needed**:
By using the Golang plugin mechanism https://pkg.go.dev/plugin,  people can develop their own custom plugins. some less generic functions do not necessarily have to be added to the core code, in order to prevent the core functions from becoming too bloated.  meanwhile, some plugins can be designed more for internal use and internal maintenance if needs

petersutter:
Why would we require the additional complexity of a plugin mechanism? Why isn't it sufficient to call gardenctl and consume it's output in your preferred format (json/yaml) if necessary?

tedteng:
in fact, I am considering using extension functions to implement some customized requirements. such as implementing a simple command to combine and display the output of cloud provider information during troubleshooting for personal purposes, instead of using `aws ec2 xxx`  check one by one. or https://github.com/gardener/gardenctl-v2/pull/162 become plugins instead of in the main code. 
Anyways, as you mentioned might output the preferred format. or develop some small tools in my local is also an option.

PR #300, opened 2023-05-12
plugin feature

tedteng:
introduce plugins feature 

```
(N/A:N/A)➜  plugins ls -lrt
total 3840
-rw-r--r-- 1 xxxxx 3930736 May 12 12:56 example.so
```
```
go run main.go
...
Usage:
  gardenctl [command]

Available Commands:
  ....
  example      A brief description of your command 
 
```

```
go run main.go example
Test called
```

Fixes https://github.com/gardener/gardenctl-v2/issues/299

```
var NewCmd = &cobra.Command{  //must be `NewCmd` name
```
~~when "hidden:false",  `make gen-markdown` will generate all documentation, including the plugin's description, because https://github.com/gardener/gardenctl-v2/compare/master...tedteng:gardenctl-v2:plugins?expand=1#diff-f36fcc143c1388c7c89b28b165cfb74d7f4c00fd6e58667a5594437f13180436R127-R129 and plugins(example.so) use `cobra` library as well.~~ 

**Release note**:

```feature user
introduce plugins feature
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,3 +185,29 @@ Establish an SSH connection to a Shoot cluster's node.
 ```bash
 gardenctl ssh my-node
 ```
+### Plugins
+To use [plugins](https://pkg.go.dev/plugin) , you have to make a `plugins` directory under `.garden`, and place the plugin files(with a .so extension) within that directory.
+
+example plugin, 
+```
+package main
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+var NewCmd = &cobra.Command{  //must be `NewCmd` name
+	Use:   "example",
+	Short: "A brief description of your command ",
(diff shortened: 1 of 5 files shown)


petersutter:
Hi @tedteng, are you aware of all the downsides that come with the golang plugins? If not I recommend checking out e.g. https://www.reddit.com/r/golang/comments/b6h8qq/is_anyone_actually_using_go_plugins/. Also (it's mentioned somewhere in the comments in this thread) windows is not supported.

tedteng:
ah, didn't aware of it. my initial idea was to implement some less common features using plugins, if they proved useful, then migrate them into the main code if needed.  the experiment in my local shows that the lower golang version main code cannot recognize higher golang version plugins and only works with higher golang version in main code with low/same version of plugins

PR #309, opened 2023-05-25
`ssh`: add flag to confirm access restrictions

petersutter:
This PR adds a new flag called `--confirm-access-restriction` to the `ssh` command. It functions similarly to the now deprecated `--force` flag of the `provider-env` command.

The `provider-env` command has also been updated to include the `--confirm-access-restriction` flag, which replaces the deprecated `--force` flag. The `--force` flag will eventually be removed in a future version.

Fixes #

**Release note**:

```feature user
`ssh`: Added flag `--confirm-access-restriction`. Using this flag bypasses the need for confirmation of any access restrictions. Set this flag only if you are fully aware of the access restrictions.
```

```other user
`provider-env`: ⚠️ The `--force` is now deprecated. Use `--confirm-access-restriction` flag instead. The `--force` flag will eventually be removed in a future version.
```

### DIFF
--- a/docs/help/gardenctl_provider-env.md
+++ b/docs/help/gardenctl_provider-env.md
@@ -34,15 +34,16 @@ gardenctl provider-env [flags]
 ### Options
 
 ```
-      --control-plane    target control plane of shoot, use together with shoot argument
-  -f, --force            Generate the script even if there are access restrictions to be confirmed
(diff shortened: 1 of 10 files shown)


holgerkoser (approved):
/lgtm

Issue #325, opened 2023-07-04
Enable `sshd` service when using `ssh` on a specific node


AleksandarSavchev:
**What would you like to be added**:
Currently it is not possible to use `gardenctl ssh` on nodes when the cluster has `provider.workersSettings.sshAccess.enabled` set to `false`. I would like to enable `ssh` access when using `gardenctl ssh` on a specific node if it had been disabled and disable it again when the `ssh` session ends.

**Why is this needed**:
Improve debugging of issues when cluster has `provider.workersSettings.sshAccess.enabled` set to `false`.

petersutter:
I believe that gardenctl should not manipulate this setting and attempt to restore it afterwards, as suggested in #331. Please refer to my [comment](https://github.com/gardener/gardenctl-v2/pull/331#issuecomment-1684086895) for reasons why I consider this to be a poor approach.

Instead, I suggest to extend `Shoot.spec.provider.workerSettings.sshAcces` and rather have something like
1.
```yaml
    workersSettings:
      sshAccess:
        enabled: false
        # policy: # has no effect if set 
```
2.
```yaml
    workersSettings:
      sshAccess:
        enabled: true
        policy: 'ondemand' # allows Bastions to be created. sshd will be running on the worker nodes as long as there are Bastion resources
```
3.
```yaml
    workersSettings:
      sshAccess:
        enabled: true
        policy: 'always' # regardless if there are Bastion resources, the sshd is running on the worker nodes
```

We currently support options 1 and 3. The newly introduced feature is option 2. If a Bastion is created, the sshd will be initiated and subsequently terminated once the last bastion resource has been deleted.

This eliminates the need for gardentl to manipulate the sshAccess settings on the Shoot and attempt to restore the initial enabled state. Moreover, the shoot owner retains control of the setting. This means that if it's disabled, debugging is not possible unless it's explicitly enabled.

@AleksandarSavchev @rfranzke @dimityrmirchev (as you were involved in #7188), WDYT?

AleksandarSavchev:
I agree with you that `gardenctl` should not modify the setting.

I approve your suggested approach and would like to point out that when option 2 is used it would work the same way as option 1. When a Bastion is created for a node in a cluster using option 2 the following steps should happen:

- `sshd-ensurer` service ( this service keeps sshd disabled) should be stopped for the specified node.
- `sshd` service should be enabled and started
- a `SSHPublicKey` should be created. 

[...]

PR #331, opened 2023-08-17
ssh when sshAccess disable

tedteng:
for issue https://github.com/gardener/gardenctl-v2/issues/325,  however, check with MCM colleagues, don't find the solution to Enable/disable sshd service on a specific node.   it seems the only way is enabling sshaccess then generate ssh key and enable sshd service affect all the worker nodes.

the solution here is 
1) to patch shoot workersSettings.sshAccess when shoot `workersSettings.sshAccess` is `false`, 
2) waiting ssh key to generate 
3) then apply bastion annotation `["WorkerSSHD"] = false` , 
will add extra logic in `delete` phase in the bastion extension repo individually and verify it.  in `delete` phase will handle revert back code to patch shoot `workersSettings.sshAccess ` back to original value `false` when detecting annotation `["WorkerSSHD"] = false` in case any issue `gardenctl` break or `keep bastion` case  

Fixes #

**Release note**:

```feature operator
gardenctl ssh worker node when shoot disable sshAccess 
```


tedteng:
@petersutter welcome any feedback for the solution discussion.  I'm also verifying and looking for a solution. 

> You suggested patching the shoot setting with gardenctl and then reverting it back to its previous value using gardenctl once a signal is received. However, if the ssh command is run multiple times for the same shoot, an incorrect value could be restored.
> You also proposed setting an annotation on the bastion to store the "initial" enabled state. This would potentially allow the extension to patch the shoot to restore this value if it was not restored by gardenctl for any reason. However, this approach has the same issue: it's unclear what the real initial value is if the command was run multiple times.

it should only have a value stored once.  user A executed the gardenctl , then shoot patch to `true`.  start from this pointer, the next user will ssh directly due to `workersSettings.sshAccess` being `true` already, there is no difference like before using ssh. 

From the bastion object pointer,  it should only have one bastion object with annotation to store the `initial` value (["WorkerSSHD"] = "false") at the beginning when user A executed the gardenctl and shoot ``workersSettings.sshAccess` is `false`.  

The next/new user will create a new bastion object same as the normal ssh process.   then only the first bastion with ["WorkerSSHD"] = "false" annotation, will process the shoot reverting process what bastion object `delete` phase triggered by gardenctl or keep-life end 


The purpose for `reverting it back to its previous value using gardenctl once a signal is received` is only in case the user cancel(Ctrl+C) the session during the shoot `workersSettings.SSHAccess.Enabled` patch before bastion created at that time.  just for double safety, I involved another controller (the bastion extension controller) to revert the value back during the bastion `delete` phase. in this way always make sure the value is revetted back the initial value. 


> Additionally, this approach would require another controller (the bastion extension controller) to patch the shoot resource, but does the user even have permission to change this setting on the shoot?

I have the same concern too and will experiment with it. 


> Also, if ssh access is disabled, do not silently enable it. The user/operator should be notified and asked if it should be enabled and there should be a flag to bypass the need for confirmation, similar to the `confirm-access-restriction` flag

sure, I also want to have a similar flag or reminder when executed gardenctl

tedteng:
close this PR based on the discussion the new feature might be implemented in g/g side instead of triggering from gardenctl

PR #348, opened 2023-10-13
Increase rsa key size to 3072

dimityrmirchev:
~~Gardener uses 4096 bits to generate ssh keys. Let's use the same sizing here as well.~~ See https://github.com/gardener/gardenctl-v2/pull/348#issuecomment-1761317716

Similar to https://github.com/gardener/gardener/pull/8635

https://github.com/gardener/gardener/blob/3e9f4819aa8ab5badacf76f8bb7c1dae31192bf3/pkg/operation/botanist/secrets.go#L255
Fixes #

**Release note**:

```other operator
`ssh` now uses 3072 bit keys
```

### DIFF
--- a/hack/test-integration.env
+++ b/hack/test-integration.env
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+export GOMEGA_DEFAULT_EVENTUALLY_TIMEOUT=5s
+export GOMEGA_DEFAULT_EVENTUALLY_POLLING_INTERVAL=200ms
--- a/hack/test-integration.env
+++ b/hack/test-integration.env
@@ -1,6 +1,0 @@
-# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
-#
-# SPDX-License-Identifier: Apache-2.0
-
-export GOMEGA_DEFAULT_EVENTUALLY_TIMEOUT=5s
-export GOMEGA_DEFAULT_EVENTUALLY_POLLING_INTERVAL=200ms
--- a/pkg/cmd/ssh/ssh_suite_test.go
+++ b/pkg/cmd/ssh/ssh_suite_test.go
@@ -27,7 +27,6 @@ func init() {
 
 func TestCommand(t *testing.T) {
 	SetDefaultEventuallyTimeout(5 * time.Second)
-	SetDefaultEventuallyPollingInterval(200 * time.Millisecond)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "SSH Command Test Suite")
 }
--- a/pkg/cmd/ssh/ssh_suite_test.go
+++ b/pkg/cmd/ssh/ssh_suite_test.go
@@ -9,6 +9,7 @@ package ssh_test
 import (
 	"os"
 	"testing"
+	"time"
 
 	gardencorev1beta1 "github.com/gardener/gardener/pkg/apis/core/v1beta1"
 	operationsv1alpha1 "github.com/gardener/gardener/pkg/apis/operations/v1alpha1"
@@ -25,6 +26,8 @@ func init() {
 }
 
 func TestCommand(t *testing.T) {
+	SetDefaultEventuallyTimeout(5 * time.Second)
+	SetDefaultEventuallyPollingInterval(200 * time.Millisecond)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "SSH Command Test Suite")
 }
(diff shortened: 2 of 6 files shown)


hack/test-integration.env
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2023 SAP SE or an SAP affiliate company and Gardener contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+export GOMEGA_DEFAULT_EVENTUALLY_TIMEOUT=5s
+export GOMEGA_DEFAULT_EVENTUALLY_POLLING_INTERVAL=200ms
```
petersutter, line 6: was there a special reason to overwrite the defaults?

dimityrmirchev: Yes, the default timeout of 1 sec was too short. I did some additional investigation and benchmarking. I will provide the results in a another comment.

pkg/cmd/ssh/ssh_test.go
```diff
@@ -47,32 +47,20 @@ type bastionStatusPatch func(status *operationsv1alpha1.BastionStatus)
 func waitForBastionThenPatchStatus(ctx context.Context, gardenClient client.Client, bastionName string, namespace string, patcher bastionStatusPatch) {
 	defer GinkgoRecover()
 
-	ticker := time.NewTicker(1 * time.Second)
```
petersutter, line 50: hm not sure why it was done that complicated in the first place with the ticker 🤔

dimityrmirchev:
I did some additional testing and it seems that generating a `4096` bit RSA key on a client machine will probably not be the most enjoyable experience since it takes significantly more time in comparison to a `2048` bit RSA key. I propose that we use a `3072` bit key since this also probably satisfies the recommendations from different security agencies and will be faster to generate. In a later stage we might consider replacing RSA with ECDSA which is significantly faster to generate keys due the smaller bit sizes providing the same [security strength](https://en.wikipedia.org/wiki/Security_level).

```golang

import (
	"crypto/ecdsa"
	"crypto/elliptic"
	"crypto/rand"
	"crypto/rsa"
	"testing"
)

func BenchmarkRSA2048(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_, err := rsa.GenerateKey(rand.Reader, 2048)
		if err != nil {
			b.Fatalf("error occurred")
		}
	}
}

func BenchmarkRSA3072(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_, err := rsa.GenerateKey(rand.Reader, 3072)
		if err != nil {
			b.Fatalf("error occurred")
		}
	}
}

func BenchmarkRSA4096(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_, err := rsa.GenerateKey(rand.Reader, 4096)
		if err != nil {
			b.Fatalf("error occurred")
		}
	}
}

func BenchmarkECDSA256(b *testing.B) {
	for i := 0; i < b.N; i++ {
		_, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
		if err != nil {
			b.Fatalf("error occurred")
		}
	}
}

```


```bash
go test -bench=.  -benchtime=10s
goos: darwin
goarch: arm64
pkg: github.com/test/play
BenchmarkRSA2048-12                  134          90775602 ns/op
BenchmarkRSA3072-12                   36         374908398 ns/op
BenchmarkRSA4096-12                   15        1335982369 ns/op
BenchmarkECDSA256-12             1000000             11136 ns/op
PASS
ok      github.com/test/play    72.649s
```

pkg/cmd/ssh/ssh_test.go
dimityrmirchev: Yes, I was wondering the same, because `Eventually` will block then retry by itself until the timeout is reached or the condition is met 😅

hack/test-integration.env
petersutter: but is it only failing on the CI system? otherwise I'd suggest to increase the respective timeout in code on where it usually fails or for all in the test suite, so that when you run the debugger the correct values are set. Otherwise you have to remember so set the same env in the debugger. WDYT?

pkg/cmd/ssh/ssh_suite_test.go
```diff
@@ -25,6 +26,8 @@ func init() {
 }
 
 func TestCommand(t *testing.T) {
+	SetDefaultEventuallyTimeout(5 * time.Second)
+	SetDefaultEventuallyPollingInterval(200 * time.Millisecond)
```
petersutter, line 30:
```suggestion
```
remove overwriting the default polling interval if there is no real requirement for it.

dimityrmirchev: Sure, done with https://github.com/gardener/gardenctl-v2/pull/348/commits/16b5084bfe2a8b960e04d42d5089fd146839eda5

petersutter (approved):
/lgtm

PR #366, opened 2024-01-10
Nix flake enablement for gardenctl-v2

vasu1124:

We want to enable gardenctl for the nix community.

```feature user
# with Nix (macOS, Linux, and Windows)
nix run github:gardener/gardenctl-v2 -- version

# development version
nix profile install github:gardener/gardenctl-v2

gardenctl --help
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,103 +7,114 @@ SPDX-License-Identifier: Apache-2.0
   description = "Nix flake for gardenctl-v2";
 
   inputs = {
-    # NixPkgs (nixos-23.11)
-    nixpkgs.url = "nixpkgs/nixos-23.11"; #"github:NixOS/nixpkgs/nixos-23.11";
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
   };
 
-  outputs = { self, nixpkgs, ... }:
-    let
-      pname = "gardenctl";
+  outputs = {
+    self,
+    nixpkgs,
+    ...
+  }: let
+    pname = "gardenctl";
 
-      # System types to support.
-      supportedSystems = [ "x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];
+    # System types to support.
+    supportedSystems = ["x86_64-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin"];
 
-      # Helper function to generate an attrset '{ x86_64-linux = f "x86_64-linux"; ... }'.
-      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+    # Helper function to generate an attrset '{ x86_64-linux = f "x86_64-linux"; ... }'.
+    forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
 
-      # Nixpkgs instantiated for supported system types.
-      nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
-    in
-    {
-      # Provide some binary packages for selected system types.
-      packages = forAllSystems (system:
-        let
-          pkgs = nixpkgsFor.${system};
-          inherit (pkgs) stdenv lib;
-        in
-        {
-          ${pname} = pkgs.buildGo121Module rec {
-            inherit pname self;
-            version = lib.fileContents ./VERSION;
-            splitVersion = lib.versions.splitVersion version;
-            major = if ((lib.elemAt splitVersion 0) == "v") then lib.elemAt splitVersion 1 else lib.elemAt splitVersion 0;
-            minor = if ((lib.elemAt splitVersion 0) == "v") then lib.elemAt splitVersion 2 else lib.elemAt splitVersion 1;
-            gitCommit = if (self ? rev) then self.rev else self.dirtyRev;
-            state = if (self ? rev) then "clean" else "dirty";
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -0,0 +1,109 @@
...
+SPDX-FileCopyrightText: 2021 SAP SE or an SAP affiliate company and Gardener contributors
+
+SPDX-License-Identifier: Apache-2.0
+*/
+{
+  description = "Nix flake for gardenctl-v2";
+
+  inputs = {
+    # NixPkgs (nixos-23.11)
+    nixpkgs.url = "nixpkgs/nixos-23.11"; #"github:NixOS/nixpkgs/nixos-23.11";
```
SuperSandro2000, line 11:
```suggestion
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
```

flake.nix
```diff
@@ -0,0 +1,109 @@
...
+          pkgs = nixpkgsFor.${system};
+          inherit (pkgs) stdenv lib;
+        in
+        {
+          ${pname} = pkgs.buildGo121Module rec {
+            inherit pname self;
+            version = lib.fileContents ./VERSION;
+            splitVersion = lib.versions.splitVersion version;
+            major = if ((lib.elemAt splitVersion 0) == "v") then lib.elemAt splitVersion 1 else lib.elemAt splitVersion 0;
+            minor = if ((lib.elemAt splitVersion 0) == "v") then lib.elemAt splitVersion 2 else lib.elemAt splitVersion 1;
```
SuperSandro2000, line 40: This can be replaced with lib.versions.major/minor.

flake.nix
```diff
@@ -0,0 +1,109 @@
...
+            subPackages = [
+              "/"
+            ];
+            nativeBuildInputs = [ pkgs.installShellFiles ];
+
+            postInstall = ''
+              mv $out/bin/${pname}-v2 $out/bin/${pname}
+              installShellCompletion --cmd ${pname} --zsh  <($out/bin/${pname} completion zsh)
+              installShellCompletion --cmd ${pname} --bash <($out/bin/${pname} completion bash)
+              installShellCompletion --cmd ${pname} --fish <($out/bin/${pname} completion fish)
```
SuperSandro2000, line 71:
```suggestion
              installShellCompletion --cmd ${pname} \
                --zsh  <($out/bin/${pname} completion zsh) \
                --bash <($out/bin/${pname} completion bash) \
                --fish <($out/bin/${pname} completion fish)
```

flake.nix
```diff
@@ -0,0 +1,109 @@
...
+        let 
+          pkgs = nixpkgsFor.${system};
+        in
+        {
+          default = pkgs.mkShell {
+            buildInputs = with pkgs; [ 
+              go_1_21 
+              gopls 
+              gotools 
+              go-tools
```
SuperSandro2000, line 100:
```suggestion
              gotools # goimports
              go-tools # staticcheck
```

flake.nix
```diff
@@ -0,0 +1,109 @@
...
+              gnumake
+            ];
+          };
+        });
+
+      # The default package for 'nix build'
+      defaultPackage = forAllSystems (system: self.packages.${system}.${pname});
+    };
+}
```
SuperSandro2000, line 109: Missing a final newline

SuperSandro2000 (commented):
I've left the explanations for the other things in https://github.com/open-component-model/ocm/pull/625#pullrequestreview-1817811477 since I didn't want to copy paste them here.

flake.nix
vasu1124: please show me how. there is a "v" in from of the semver string.

vasu1124:
> Would it be possible to also add an overlay to the flake?

Give me some time to check the pro & cons. I first need the maintainers to merge this PR (but they insist that the vendorhash has to be automated as well ...)

>add gardenlogin to the inputs of this flake

Will check as well. I currently like that this dependency is not made implicit (meaning, the user has to explicitly flake download each tool).

>When trying out this flake I additionally encountered the following error on shell startup:

If you setup `nix` properly, the `installShellCompletion` of the derivation automagically includes the `source <(tool-of-choice completion $SHELL)`, and those dotfiles/directories are made read-only. I think you have to seek out how to run your environment side by side with `nix`, the issue is that you are trying to overwrite `nix`.

Kumm-Kai:
I 100% agree that the `vendorHash` should be updated automatically (at the very least CI should check if the flake can still build)... Otherwise, it's pretty much useless 🙃

@vasu1124 

> If you setup nix properly, the installShellCompletion of the derivation automagically includes the source <(tool-of-choice completion $SHELL), and those dotfiles/directories are made read-only. I think you have to seek out how to run your environment side by side with nix, the issue is that you are trying to overwrite nix.

I'm aware that most tools (installed with nix) are "installing" the completion scripts into a location I can use by adding it to `$fpath` (which I already do). However, the `source <(gardenctl rc zsh)` call in my `.zshrc` (documented [here](https://github.com/gardener/gardenctl-v2/blob/master/docs/help/gardenctl_rc.md)) is additionally used to create aliases and "start" the gardenctl environment.

I could specify everything on my own, instead of using `source <(gardenctl rc zsh)`, but IMO the script ideally shouldn't create & save the completion scripts on every shell startup / call of the script.

PR #449, opened 2024-06-04
Improve "gardenctl target view" formatting

rrhubenov:
Improve the format of the "gardenctl target view" subcommand output for better readability.

Before:
```
garden:"local", project:"local", shoot:"local"
```

After:
```
garden: local
project: local
shoot: local
```
Fixes #

**Release note**:

```feature user
NONE
```

### DIFF
--- a/pkg/env/templates/rc.tmpl
+++ b/pkg/env/templates/rc.tmpl
@@ -3,12 +3,12 @@ if [ -z "$GCTL_SESSION_ID" ] && [ -z "$TERM_SESSION_ID" ]; then
   export GCTL_SESSION_ID=$(uuidgen)
 fi
 alias {{.prefix}}=gardenctl
-alias {{.prefix}}tv='gardenctl target view -o yaml'
+alias {{.prefix}}tv='gardenctl target view'
 alias {{.prefix}}tc='gardenctl target control-plane'
 alias {{.prefix}}tc-='gardenctl target unset control-plane'
 alias {{.prefix}}k='eval "$(gardenctl kubectl-env {{.shell}})"'
 alias {{.prefix}}p='eval "$(gardenctl provider-env {{.shell}})"'
-alias {{.prefix}}cv='gardenctl config view -o yaml'
+alias {{.prefix}}cv='gardenctl config view'
 {{if not .noCompletion -}}
 source <(gardenctl completion {{.shell}})
 complete -o default -F __start_gardenctl {{.prefix}}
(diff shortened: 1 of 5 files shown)


pkg/target/target.go
```diff
@@ -161,30 +161,30 @@ func (t *targetImpl) String() string {
...
 		steps = append(steps, "control plane targeted")
 	}
 
 	if len(steps) == 0 {
-		return "<empty>"
+		return "<empty>\n"
 	}
 
-	return strings.Join(steps, ", ")
+	return strings.Join(steps, "\n") + "\n"
```
petersutter, line 187: as alternative, we could set the `yaml` output as default, as we do [here](https://github.com/gardener/gardenctl-v2/blob/d889ebb88b103b3cb7695fc66682907af06e5706/pkg/cmd/config/view.go#L53-L55) and get rid of the `String()` implementation as there is currently not much difference
We already have the `yaml` output as default for the `gtv` alias https://github.com/gardener/gardenctl-v2/blob/d889ebb88b103b3cb7695fc66682907af06e5706/pkg/env/templates/rc.tmpl#L11
We could then also omit the `-o yaml` addition on the alias

rrhubenov: Hi Peter, thanks for the suggestion! 
I agree that setting yaml as the default output format is better. I've reworked the 'target view' command to be similar to 'config view' and I've removed the I've removed the `String()` method.

pkg/env/templates/rc.tmpl
```diff
@@ -23,12 +23,12 @@ if [ -z "$GCTL_SESSION_ID" ] && [ -z "$TERM_SESSION_ID" ]; then
   export GCTL_SESSION_ID=$(uuidgen)
 fi
 alias {{.prefix}}=gardenctl
-alias {{.prefix}}tv='gardenctl target view -o yaml'
```
petersutter, line 26: the templates for the other shells also have to be adapted

```bash
cat pkg/env/templates/rc.tmpl | grep -n 'gardenctl target view -o yaml'
6:alias {{.prefix}}tv='gardenctl target view -o yaml'
58:alias {{.prefix}}tv='gardenctl target view -o yaml'
79:  gardenctl target view -o yaml
```

pkg/env/templates/rc.tmpl
```diff
@@ -23,12 +23,12 @@ if [ -z "$GCTL_SESSION_ID" ] && [ -z "$TERM_SESSION_ID" ]; then
...
+alias {{.prefix}}tv='gardenctl target view'
 alias {{.prefix}}tc='gardenctl target control-plane'
 alias {{.prefix}}tc-='gardenctl target unset control-plane'
 alias {{.prefix}}k='eval "$(gardenctl kubectl-env {{.shell}})"'
 alias {{.prefix}}p='eval "$(gardenctl provider-env {{.shell}})"'
-alias {{.prefix}}cv='gardenctl config view -o yaml'
```
petersutter, line 31: there are still some occurences here as well

```bash
cat pkg/env/templates/rc.tmpl | grep -n 'gardenctl config view -o yaml'
11:alias {{.prefix}}cv='gardenctl config view -o yaml'
63:alias {{.prefix}}cv='gardenctl config view -o yaml'
99:  gardenctl config view -o yaml
```

petersutter (approved):
/lgtm, though I'd not rename it to `TargetViewOptions` as `target.ViewOptions` is clear and concise

grolu (approved):
/lgtm

holgerkoser (approved):
/lgtm

PR #469, opened 2024-10-24
Add binaries as release assets to component descriptor

8R0WNI3:
To model the built gardenctl-v2 binaries as delivery relevant artefacts, they should be part of the component descriptor's resources. Since the binaries are already configured as output of the `build` step, this output can be used to create additional release assets.

Fixes #

This changes requires https://github.com/gardener/cc-utils/pull/1065 (i.e. cc-utils version >= `1.2535.0`) because of the newly added `artefact_extra_id` property.

**Release note**:

```feature user

```

### DIFF
--- a/.ci/build
+++ b/.ci/build
@@ -22,6 +22,8 @@ fi
 ###############################################################################
 
 "${MAIN_REPO_DIR}"/hack/build-linux-amd64.sh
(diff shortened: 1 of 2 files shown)


ccwienk (approved):

petersutter (approved):
/lgtm, thanks!

PR #470, opened 2024-10-25
Introduce `gosec` for Static Application Security Testing (SAST)

petersutter:
This PR introduces `gosec` for Static Application Security Testing at Gardener and should replace other code scanners.

It uses the default ruleset of `gosec` with two exceptions.
1. Scanning generated code is disabled. The Kubernetes code-gen creates hundreds of low priority findings for unsafe pointers (G103). The generated code has proven that it is working and it is not possible to address those findings.
2. The `./hack` folder is excluded, because `gosec` does not support nested go modules (like logcheck) and there is no productive code in this folder anyway. 

All findings based on these settings have been mitigated in this PR.

There are two new `make` targets.
- `make sast` is supposed to be called in unit tests. It logs all finding to console and fails in case of errors.
- `make sast-report` can be used in build pipelines. It fails on errors too. Additionally, it creates a report in sarif format to `./gosec-report.sarif` file. This report also tracks suppressed findings.

Fixes #

**Release note**:

```feature developer
`gosec` was introduced for Static Application Security Testing (SAST).
```


holgerkoser (approved):
/lgtm

Issue #509, opened 2025-01-20
deduplicate builds / publishing pipelines


ccwienk:
# Context / Motivation

Buildresults from this repository are built and published twice:
- using concourse-pipeline / publishing as OCM-Component + github-release-assets [ref](https://github.com/gardener/gardenctl-v2/blob/master/.ci/pipeline_definitions#L38-L90)
- using GitHub-Actions/workflows to publish to brew / chocolatery [ref](https://github.com/gardener/gardenctl-v2/blob/master/.github/workflows/update-gardenctl-v2.yaml)

Having duplicate pipelines (which both build the same build-results) has the disadvantage of publishing different binaries (even for same sources, builds will typically yield slightly different binaries, as e.g. timestamps will affect output).

Considering there is an ongoing activity to migrate from Concourse -> GitHub-Actions, an obvious choice might be to extend existing GHA-Workflow to also create/publish GitHub-Releases and OCM-Component-Descriptors (there are some reusable github-actions / workflows at [cc-utils](https://github.com/gardener/cc-utils/tree/master/.github) - currently still WIP).

petersutter:
As first measurement we have disabled the binary upload from the concurse pipeline [ref](https://github.com/gardener/gardenctl-v2/issues/509)

petersutter:
fixed with https://github.com/gardener/gardenctl-v2/pull/598

PR #511, opened 2025-01-21
chore: Disable binary upload

petersutter:
We should transition to reusing the binaries generated by the CI pipeline instead of building them separately in the Action (or vice versa, as described in https://github.com/gardener/gardenctl-v2/issues/509). Currently, we have disabled the binary upload from the CI pipeline, but this should be re-enabled once we implement the proper reuse strategy.

Fixes #

https://github.com/gardener/cc-utils/pull/1120

**Release note**:

```feature user

```


grolu (approved):
/lgtm

PR #565, opened 2025-05-09
Support ssh to IPv6 addresses.

axel7born:
When connecting to nodes that have IPv6 addresses, the address must be in brackets in order to not confuse the colons in the address with  the port specification.

Fixes #

**Release note**:

```feature user
ssh to nodes that have IPv6 addresses is now supported. 
```


petersutter (approved):
/lgtm, thanks!

PR #574, opened 2025-05-26
Add additional GCP service account validation

petersutter:

This PR adds additional security validation for GCP service account credentials used in the `gardenctl provider-env` command. It validates that service account JSON fields like `universe_domain`, `token_uri`, `auth_uri`, and certificate URLs only contain trusted values matching predefined patterns.

The command works as before without any configuration changes, using secure defaults for standard Google Cloud endpoints. Additional patterns can be configured if needed via the config file (`provider.gcp.allowedPatterns`) or `--gcp-allowed-patterns` CLI flag.

Fixes #

**Release note**:

```feature user
Added additional GCP service account validation for `provider-env` command
```


grolu (approved):
/lgtm

docs/help/gardenctl_provider-env.md
```diff
@@ -27,23 +27,50 @@ for the respective provider in the "templates" folder of the gardenctl home dire
...
-      --garden string                target the given garden cluster
-  -h, --help                         help for provider-env
-  -o, --output string                One of 'yaml' or 'json'.
-      --project string               target the given project
-      --seed string                  target the given seed cluster
-      --shoot string                 target the given shoot cluster
-  -u, --unset                        Generate the script to unset the cloud provider CLI environment variables and logout for 
+  -y, --confirm-access-restriction     Confirm any access restrictions. Set this flag only if you are completely aware of the access restrictions.
+      --control-plane                  target control plane of shoot, use together with shoot argument
+  -f, --force                          Deprecated. Use --confirm-access-restriction instead. Generate the script even if there are access restrictions to be confirmed.
```
klocke-io, line 65: Nice, making it more explicit 💪

pkg/cmd/providerenv/gcp.go
```diff
@@ -0,0 +1,182 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// EncodedClientEmailPlaceholder is a placeholder string used in URL patterns to be replaced with the URL-encoded client email from the service account credentials.
+const EncodedClientEmailPlaceholder = "{encoded_client_email}"
+
+// defaultAllowedPatterns returns the default allowed patterns for GCP service account fields.
+func defaultAllowedPatterns() map[string][]string {
+	return map[string][]string{
```
klocke-io, line 25: Could be extracted, especially if we roll out this approach to more providers. I think it would be nice having them in a dedicated file / folder. But I have no strong feelings here. I also see a value in keeping it simple and not optimising for "potential" future growth

klocke-io (approved):
/lgtm 

Nice 🚀

pkg/cmd/providerenv/gcp.go
petersutter: Thanks for the suggestion! I agree that extracting this could make sense if we expand to more providers with similar patterns, but for now, I'd prefer keeping it in `gcp.go` since the allowed patterns are tightly coupled to GCP's specific JSON credential structure (stored as `serviceaccount.json` in K8s secrets, with fields like `token_uri`). Other providers like AWS handle creds differently (e.g., stored as individual keys like `accessKeyID` in K8s secrets), so mixing them might complicate things without clear benefits yet. Happy to revisit if we add similar logic elsewhere.

PR #636, opened 2025-10-13
Harden `provider-env`: schema-driven credential validation and allowed patterns

petersutter:
- Enforces schema-based credential validation in `provider-env` using pluggable validators.
- OpenStack: validates `authURL` against user-configured `allowedPatterns` (via config or `--openstack-allowed-patterns` / `--openstack-allowed-uri-patterns`).
- Hardens templates
  -  avoid exporting credentials unless necessary
  - use temporary vars and unset afterward
  - expose raw Secret only via reserved `unsafeSecretData` for custom templates (unvalidated — hence potentially "unsafe"); required to keep out-of-tree provider support in gardenctl.
- Docs: adds `docs/usage/provider-env.md` and updates provider-specific guidance.

Fixes #

**Release note**:
```feature user
`provider-env` command enhancements:
- Enforce schema-driven validation for provider credentials
- Avoid exporting credentials unless needed
```

```breaking user
`provider-env` command:
- OpenStack: `provider-env` now requires configuring allowed `authURL` patterns; the command will fail for OpenStack if none are provided. See documentation: [OpenStack: Allowed authURL patterns (required)](https://github.com/gardener/gardenctl-v2/blob/master/docs/usage/provider-env.md#openstack-allowed-authurl-patterns-required)
- Templates: raw Secret data is no longer exposed at the top level in `provider-env` templates. Custom/out-of-tree templates that previously accessed raw keys must now read them from `.unsafeSecretData` *(unvalidated — hence "unsafe")*. This is currently required to keep out-of-tree provider support in gardenctl. Built-in templates must not use `.unsafeSecretData`.
```


pkg/cmd/providerenv/options.go
```diff
@@ -54,12 +57,29 @@ type options struct {
...
+	// OpenStackAllowedPatterns is a list of JSON-formatted allowed patterns for OpenStack credential config fields
+	OpenStackAllowedPatterns []string
+	// OpenStackAllowedURIPatterns is a list of simple field=uri patterns for OpenStack credential config fields
+	OpenStackAllowedURIPatterns []string
+	// MergedAllowedPatterns contains the merged allowed patterns for all providers from defaults, config, and flags
+	MergedAllowedPatterns *MergedProviderPatterns
+}
+
+// MergedProviderPatterns contains merged allowed patterns for cloud providers that support pattern-based field validation.
+// Currently, only GCP and OpenStack providers support allowed patterns.
```
klocke-io: The other provider don't have dynamic parts to it?

klocke-io (approved):
Very well structured, 
Thanks for taking the time and going the extra mile.

pkg/cmd/providerenv/options.go
petersutter: good point. I was even thinking of dropping the user-provided pattern (config and flags) for gcp. For openstack it is a necessity, as the authURL cannot be known in advance / hardcoded. Basically, if the built-in allowed patterns are not too strict / allow all valid credentials, there is no need for custom allowed patterns.

grolu (approved):
/lgtm

PR #648, opened 2025-11-05
Harden `provider-env`: separate data from code

petersutter:
This PR further hardens the `provider-env` command to separate data from code

Fixes #

**Release note**:

```breaking developer
`provider-env`: `region` and credential field values are now written to temporary files. Their file paths are exposed under `.dataFiles` both in (custom) provider templates and when using `gardenctl provider-env -o {json|yaml}` (for example, `.dataFiles.region` points to the temporary file). This change is only relevant when using custom provider templates or when invoking `gardenctl provider-env -o {json|yaml}` in custom scripts.
```

### DIFF
--- a/pkg/target/target.go
+++ b/pkg/target/target.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 
 	gardencore "github.com/gardener/gardener/pkg/apis/core"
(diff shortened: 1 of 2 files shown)


grolu (approved):
/lgtm

PR #727, opened 2026-04-22
Mention rc command in README.md

mbs-c:

The old README explained how to set the required environment variables and enable shell completion, but did not mention the easiest way to configure both (and a few nice extras): The `rc` command.

N/A

This is what I would have liked to read when I first opened the README, though.

N/A

**Release note**:
```doc user
Recommend the rc command for configuring the shell in README.md
```

## Summary by CodeRabbit

* **Documentation**
  * Added a "Recommended Shell Configuration" section with a bash example and notes for other shells; documents sourcing to handle session state, completion, and aliases. Reorganized the Shell Session section and moved an important NOTE into the new guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,27 @@ your shell completion command. Example:
 gardenctl completion bash --help
 ```
 
+### Recommended shell configuration
+
+It is possible to configure the shell session, completion and useful aliases at the same time using `gardenctl rc`.
+For more information on how to configure your preferred shell, see the included help:
+
+```sh
+gardenctl rc bash --help
+```
+
+```sh
+gardenctl rc fish --help
+```
+
+```powershell
+gardenctl rc powershell --help
+```
+
+```sh
+gardenctl rc zsh --help
+```
+
 ## Usage
 
 ### Targeting
--- a/README.md
+++ b/README.md
@@ -104,7 +104,26 @@ export GCTL_CONFIG_NAME=myconfig # without extension!
 # config is expected to be under /alternate/garden/config/dir/myconfig.yaml
 ```
 
+### Recommended Shell Configuration
+
+The easiest way to configure your shell for gardenctl is using `gardenctl rc`.
+For bash, add the following to your `~/.bashrc`:
+
+```bash
+source <(gardenctl rc bash)
+```
+
+Other shells (`zsh`, `fish`, `powershell`) are also supported. Run `gardenctl rc [shell] --help` for details.
+
+This automatically:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,6 +139,27 @@ your shell completion command. Example:
 gardenctl completion bash --help
 ```
 
+### Recommended shell configuration
```
petersutter: > > > 2. **Consider section placement**: This section appears _after_ the manual "Shell Session" and "Completion" sections, which may cause users to set things up manually before discovering the simpler approach. You might want to:
> > >    
> > >    * Move this section before "Shell Session" and "Completion", OR
> > >    * Add a brief note in the earlier sections pointing to this recommended approach



Agree with CodeRabbit. Move this section before "Shell Session" and "Completion". It's the recommended approach; users shouldn't configure things manually first only to find the simpler option below.


Also add a hint to the manual sections below: Since "Shell Session" and "Completion" become the manual alternative, add a brief note at the top of each:

  ▎ Note: If you source gardenctl rc as described above, this is already handled for you.

README.md
```diff
@@ -139,6 +139,27 @@ your shell completion command. Example:
...
+gardenctl rc fish --help
+```
+
+```powershell
+gardenctl rc powershell --help
+```
+
+```sh
+gardenctl rc zsh --help
+```
```
petersutter:
~~~suggestion
### Recommended Shell Configuration

The easiest way to configure your shell for gardenctl is using `gardenctl rc`.
For bash, add the following to your `~/.bashrc`:

```bash
source <(gardenctl rc bash)
```

Other shells (`zsh`, `fish`, `powershell`) are also supported. Run `gardenctl rc --help` for details.

This automatically:
- Sets the `GCTL_SESSION_ID` and `KUBECONFIG` environment variables, preventing the warning about a missing `KUBECONFIG` when running `gardenctl target`
- Loads shell completion
- Provides useful aliases (prefixed with `g` by default)
~~~

petersutter (commented):
thanks for adding a section for the rc command, this is something that I had since long as TODO :)
I also added a bit more context / motivation for the user, see my suggestions

README.md
mbs-c: Thanks for the review, I'll apply the suggestions. One question, though: Shouldn't we rather suggest to run "gardenctl rc __[shell]__ --help" for more details? The general help does not provide guidance for configuring the profile.

petersutter: yes good point, we should do that instead

petersutter (commented):
/lgtm

thanks!

grolu:
/lgtm
/approve

nice!

PR #735, opened 2026-05-07
feat: Allow configuration of Kubeconfig access level (config + flags)

### LINKED ISSUE #71: Allow admins to easily use a viewer kubeconfig by default
**How to categorize this topic?**

/area security
/area ops-productivity
/kind enhancement

/label teamsize/small

**What is the topic about?**:
Using gardenctl, a admin always gets a adminkubeconfig when connecting to shoots or seeds. It would be great, if I can configure in my gardenctl something like this:

```yaml
gardens:
- name: prd
  kubeconfig: /secret/path
  config:
    useViewerKubeconfig: always|shoots|managedseeds|never
```

And then also introduce a flag like `--admin` or `--viewer` to overwrite this for any `gardenctl target` command.

jamand:
**How to categorize this PR?**

/area ops-productivity
/area security
/kind enhancement

Implements admin/viewer/auto kubeconfig access-level support in `gardenctl`, so admins working on production gardens can default to a read-only kubeconfig and be confident they cannot accidentally break things.

Added new fields to the gardenctl config:
```yaml
gardens:
- identity: prd-garden
  kubeconfig: ~/path/to/kubeconfig.yaml
  kubeconfigAccessLevelDefaults:
    shoots: viewer        # admin (default) | viewer | auto
    managedSeeds: admin   # admin (default) | viewer | auto
```

Added a `--access-level` flag to relevant commands (`target`, `kubeconfig`) to allow overwriting the setting in the config for the currently targeted `Garden`.

Added `--default-shoot-access-level` and `--default-managed-seed-access-level` flags to the `gardenctl config set-garden` command to set the fields in the config. When `set-garden` changes the access levels, `gardenctl` prints a hint that existing session kubeconfigs are not regenerated and that `gardenctl target` may need to be re-run.

The flag value is plumbed through `gardenlogin`'s existing `--access-level` flag, which already supports admin/viewer/auto end-to-end. For non-managed seeds (and the garden cluster) `gardenlogin`'s access-level mechanism is unavailable, so an explicit viewer request returns an error; `admin` and `auto` use the static seed-login kubeconfig as-is.

**Behavior**:
- Defers to gardenlogin's own default when nothing is set.
- Scope by cluster role, not target path (that is if doing `target shoot xy` and this `Shoot` is backend by a `ManagedSeed` you would get the permissions configured under `kubeconfigAccessLevelDefaults.managedSeeds`.
 
Fixes gardener/hackathon#71

**Release note**:

```feature operator
`gardenctl` now supports requesting a read-only viewer kubeconfig instead of admin via the `--access-level` flag (`admin` / `viewer` / `auto`) and the per-garden `kubeconfigAccessLevelDefaults.{shoots,managedSeeds}` config block (also editable via `gardenctl config set-garden`).
```

## Summary by CodeRabbit

* **New Features**
  * Add kubeconfig access-level support (admin/viewer/auto) with --access-level and --admin/--viewer shorthands across target, kubeconfig, kubectl-env and related commands.
  * Add per-garden defaults flags (--default-shoot-access-level, --default-seed-access-level) in config set-garden; changing defaults prints a hint to re-run target.
  * Commands now report the effective access level (and append it to target success messages when applicable); kubectl-env warns when symlink mode makes --access-level ineffective.

* **Documentation**
  * README, configuration guide, and CLI help updated with examples, flag behavior, validation rules, and verification guidance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ gardens:
 # name: my-name # An alternative, unique garden name for targeting
 # context: different-context # Overrides the current-context of the garden cluster kubeconfig
 # patterns: ~ # List of regex patterns for pattern targeting
-# defaultKubeconfigAccessLevel: # Default access level requested per target scope
+# kubeconfigAccessLevelDefaults: # Default access level requested per target scope
 #   shoots: viewer              # admin (default) | viewer | auto
 #   managedSeeds: admin         # admin (default) | viewer | auto
 ```
@@ -107,10 +107,10 @@ For shoots and managed seeds, `gardenctl` requests a kubeconfig from `gardenlogi
 - `viewer` - read-only credentials with no access to encrypted resources (e.g. Secrets).
 - `auto` - try admin first; fall back to viewer if admin is denied. Convenient when you don't know your RBAC up front; for users who *do* have admin, this is effectively the same as `admin`.
 
-Set per-scope defaults per garden under `defaultKubeconfigAccessLevel`. The `shoots` and `managedSeeds` fields are independent - for example, an admin who wants least-privilege access to customer shoots while still keeping admin access for seed-level debugging would set:
+Set per-scope defaults per garden under `kubeconfigAccessLevelDefaults`. The `shoots` and `managedSeeds` fields are independent - for example, an admin who wants least-privilege access to customer shoots while still keeping admin access for seed-level debugging would set:
 
 ```yaml
-defaultKubeconfigAccessLevel:
+kubeconfigAccessLevelDefaults:
   shoots: viewer
   managedSeeds: admin
 ```
--- a/README.md
+++ b/README.md
@@ -115,15 +115,18 @@ kubeconfigAccessLevelDefaults:
   managedSeeds: admin
 ```
 
-Override per invocation with `--kubeconfig-access-level` (or the `--admin` / `--viewer` shorthands), available on commands that produce a kubeconfig (`target`, `kubeconfig`):
+Override per invocation with `--access-level` (or the `--admin` / `--viewer` shorthands), available on commands that produce a kubeconfig (`target`, `kubeconfig`):
 
 ```bash
-gardenctl kubeconfig --admin                                    # one-off escalation
-gardenctl target --shoot my-shoot --viewer                      # one-off de-escalation
-gardenctl kubeconfig --kubeconfig-access-level=auto             # explicit auto (no shorthand)
+gardenctl kubeconfig --admin                       # one-off escalation
+gardenctl target --shoot my-shoot --viewer         # one-off de-escalation
+gardenctl kubeconfig --access-level=auto           # explicit auto (no shorthand)
 ```
 
 The shorthands and the full form are mutually exclusive — pass at most one.
+
+> [!NOTE]
+> Flag overrides apply only to the invocation they're passed to. They are not persisted on disk, so a subsequent `gardenctl kubeconfig` (without a flag) re-resolves from the per-garden config defaults, not from the flag you passed to an earlier `gardenctl target`. For persistent changes set them via `gardenctl config set-garden`. You may also repeat the flags on `gardenctl kubeconfig`.
 
 The same defaults can be set via the CLI rather than hand-editing the YAML:
 
@@ -146,7 +149,7 @@ The access level only applies to **shoots and managed seeds**, where Gardener pr
 When `gardenctl` explicitly chose an access level (via flag or per-garden config), it surfaces it inline after `target` and on stderr after `kubeconfig`:
 
 ```console
-$ gardenctl target shoot prod-cluster --kubeconfig-access-level=viewer
+$ gardenctl target shoot prod-cluster --access-level=viewer
 Successfully targeted shoot "prod-cluster" (access level: viewer)
 ```
 
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ defaultKubeconfigAccessLevel:
   managedSeeds: admin
 ```
 
-Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`, `kubectl-env`):
+Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`):
 
 ```bash
 gardenctl kubeconfig --kubeconfig-access-level=admin            # one-off escalation
--- a/README.md
+++ b/README.md
@@ -140,15 +140,16 @@ The access level only applies to **shoots and managed seeds**, where Gardener pr
 
 #### Verifying which access level is in use
 
-`gardenctl` prints the access level after `target` and `kubeconfig` (stderr):
+When `gardenctl` explicitly chose an access level (via flag or per-garden config), it surfaces it inline after `target` and on stderr after `kubeconfig`:
 
-```
-$ gardenctl target shoot prod-cluster
-Successfully targeted shoot "prod-cluster"
-Kubeconfig access level: viewer
+```console
+$ gardenctl target shoot prod-cluster --kubeconfig-access-level=viewer
+Successfully targeted shoot "prod-cluster" (access level: viewer)
 ```
 
-For an explicit check, the access level is also embedded as a `--access-level=…` argument in the produced kubeconfig's exec plugin section:
+When neither a flag nor a per-garden config is set, `gardenctl` stays silent and lets `gardenlogin`'s own default apply (currently `auto`).
+
+For an explicit check, the access level is also embedded as a `--access-level=…` argument in the produced kubeconfig's exec plugin section (omitted when `gardenctl` deferred to `gardenlogin`'s default):
 
 ```bash
 gardenctl kubeconfig | grep -- '--access-level='
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ defaultKubeconfigAccessLevel:
   managedSeeds: admin
 ```
 
-Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`, `kubectl-env`, `ssh`, `ssh-patch`):
+Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`, `kubectl-env`):
 
 ```bash
 gardenctl kubeconfig --kubeconfig-access-level=admin            # one-off escalation
--- a/README.md
+++ b/README.md
@@ -115,12 +115,15 @@ kubeconfigAccessLevelDefaults:
   managedSeeds: admin
 ```
 
-Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`):
+Override per invocation with `--kubeconfig-access-level` (or the `--admin` / `--viewer` shorthands), available on commands that produce a kubeconfig (`target`, `kubeconfig`):
 
 ```bash
-gardenctl kubeconfig --kubeconfig-access-level=admin            # one-off escalation
-gardenctl target --shoot my-shoot --kubeconfig-access-level=viewer
+gardenctl kubeconfig --admin                                    # one-off escalation
+gardenctl target --shoot my-shoot --viewer                      # one-off de-escalation
+gardenctl kubeconfig --kubeconfig-access-level=auto             # explicit auto (no shorthand)
 ```
+
+The shorthands and the full form are mutually exclusive — pass at most one.
 
 The same defaults can be set via the CLI rather than hand-editing the YAML:
 
--- a/README.md
+++ b/README.md
@@ -90,11 +90,69 @@ gardens:
 # name: my-name # An alternative, unique garden name for targeting
 # context: different-context # Overrides the current-context of the garden cluster kubeconfig
 # patterns: ~ # List of regex patterns for pattern targeting
+# defaultKubeconfigAccessLevel: # Default access level requested per target scope
+#   shoots: viewer              # admin (default) | viewer | auto
+#   managedSeeds: admin         # admin (default) | viewer | auto
 ```
 
 > [!NOTE]
 > - To use the kubeconfigs for **shoot clusters** provided by `gardenctl`, you need to have [gardenlogin](https://github.com/gardener/gardenlogin) installed as a `kubectl` auth plugin.
 > - If your **garden cluster** kubeconfig uses OIDC authentication, ensure that you have the [kubelogin](https://github.com/int128/kubelogin) `kubectl` auth plugin installed.
+
+### Kubeconfig Access Level
+
+For shoots and managed seeds, `gardenctl` requests a kubeconfig from `gardenlogin` with a configurable access level:
+
+- `admin` (default) - full cluster-admin credentials.
+- `viewer` - read-only credentials with no access to encrypted resources (e.g. Secrets).
+- `auto` - try admin first; fall back to viewer if admin is denied. Convenient when you don't know your RBAC up front; for users who *do* have admin, this is effectively the same as `admin`.
+
+Set per-scope defaults per garden under `defaultKubeconfigAccessLevel`. The `shoots` and `managedSeeds` fields are independent - for example, an admin who wants least-privilege access to customer shoots while still keeping admin access for seed-level debugging would set:
+
+```yaml
+defaultKubeconfigAccessLevel:
+  shoots: viewer
+  managedSeeds: admin
+```
+
+Override per invocation with the `--kubeconfig-access-level` flag, available on commands that produce a kubeconfig (`target`, `kubeconfig`, `kubectl-env`, `ssh`, `ssh-patch`):
+
+```bash
+gardenctl kubeconfig --kubeconfig-access-level=admin            # one-off escalation
+gardenctl target --shoot my-shoot --kubeconfig-access-level=viewer
+```
+
+The same defaults can be set via the CLI rather than hand-editing the YAML:
+
+```bash
+gardenctl config set-garden prd-garden \
+  --default-shoot-access-level viewer \
+  --default-managed-seed-access-level admin
+```
+
+The access level only applies to **shoots and managed seeds**, where Gardener provides admin/viewer kubeconfig subresources that gardenlogin requests on demand. It does not apply to:
+
+- The **garden cluster itself** - gardenctl uses the user's own kubeconfig file as configured. There is no API to "downgrade" those credentials; if you want viewer-only access to the garden, point gardenctl at a kubeconfig whose underlying RBAC is read-only.
+- **Non-managed seeds** - credentials come from a static `<name>.login` Secret in the garden cluster, which only contains admin credentials. Requesting `viewer` or `auto` for a non-managed seed (via flag or the `managedSeeds` default) returns an error.
+
+> [!NOTE]
+> `auto` returns admin whenever your RBAC allows it. If you want guaranteed read-only access, configure `viewer` explicitly.
+
+#### Verifying which access level is in use
+
+`gardenctl` prints the access level after `target` and `kubeconfig` (stderr):
+
+```
+$ gardenctl target shoot prod-cluster
+Successfully targeted shoot "prod-cluster"
+Kubeconfig access level: viewer
+```
+
+For an explicit check, the access level is also embedded as a `--access-level=…` argument in the produced kubeconfig's exec plugin section:
+
+```bash
+gardenctl kubeconfig | grep -- '--access-level='
+```
 
 ### Config Path Overwrite
 
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ gardens:
 # context: different-context # Overrides the current-context of the garden cluster kubeconfig
 # patterns: ~ # List of regex patterns for pattern targeting
 # kubeconfigAccessLevelDefaults: # Default access level requested per target scope
-#   shoots: viewer              # admin (default) | viewer | auto
-#   managedSeeds: admin         # admin (default) | viewer | auto
+#   shoots: viewer              # admin | viewer | auto (empty = gardenlogin's default)
+#   seeds: admin                # admin | viewer | auto (empty = gardenlogin's default)
 ```
 
 > [!NOTE]
@@ -101,65 +101,7 @@ gardens:
 
 ### Kubeconfig Access Level
 
-For shoots and managed seeds, `gardenctl` requests a kubeconfig from `gardenlogin` with a configurable access level:
-
-- `admin` (default) - full cluster-admin credentials.
-- `viewer` - read-only credentials with no access to encrypted resources (e.g. Secrets).
-- `auto` - try admin first; fall back to viewer if admin is denied. Convenient when you don't know your RBAC up front; for users who *do* have admin, this is effectively the same as `admin`.
-
-Set per-scope defaults per garden under `kubeconfigAccessLevelDefaults`. The `shoots` and `managedSeeds` fields are independent - for example, an admin who wants least-privilege access to customer shoots while still keeping admin access for seed-level debugging would set:
-
-```yaml
-kubeconfigAccessLevelDefaults:
-  shoots: viewer
-  managedSeeds: admin
-```
-
-Override per invocation with `--access-level` (or the `--admin` / `--viewer` shorthands), available on commands that produce a kubeconfig (`target`, `kubeconfig`):
-
-```bash
-gardenctl kubeconfig --admin                       # one-off escalation
-gardenctl target --shoot my-shoot --viewer         # one-off de-escalation
-gardenctl kubeconfig --access-level=auto           # explicit auto (no shorthand)
-```
-
-The shorthands and the full form are mutually exclusive — pass at most one.
-
-> [!NOTE]
-> Flag overrides apply only to the invocation they're passed to. They are not persisted on disk, so a subsequent `gardenctl kubeconfig` (without a flag) re-resolves from the per-garden config defaults, not from the flag you passed to an earlier `gardenctl target`. For persistent changes set them via `gardenctl config set-garden`. You may also repeat the flags on `gardenctl kubeconfig`.
-
-The same defaults can be set via the CLI rather than hand-editing the YAML:
-
-```bash
-gardenctl config set-garden prd-garden \
-  --default-shoot-access-level viewer \
-  --default-managed-seed-access-level admin
-```
-
-The access level only applies to **shoots and managed seeds**, where Gardener provides admin/viewer kubeconfig subresources that gardenlogin requests on demand. It does not apply to:
-
-- The **garden cluster itself** - gardenctl uses the user's own kubeconfig file as configured. There is no API to "downgrade" those credentials; if you want viewer-only access to the garden, point gardenctl at a kubeconfig whose underlying RBAC is read-only.
-- **Non-managed seeds** - credentials come from a static `<name>.login` Secret in the garden cluster, which only contains admin credentials. Requesting `viewer` or `auto` for a non-managed seed (via flag or the `managedSeeds` default) returns an error.
-
-> [!NOTE]
-> `auto` returns admin whenever your RBAC allows it. If you want guaranteed read-only access, configure `viewer` explicitly.
-
-#### Verifying which access level is in use
-
-When `gardenctl` explicitly chose an access level (via flag or per-garden config), it surfaces it inline after `target` and on stderr after `kubeconfig`:
-
-```console
-$ gardenctl target shoot prod-cluster --access-level=viewer
-Successfully targeted shoot "prod-cluster" (access level: viewer)
-```
-
-When neither a flag nor a per-garden config is set, `gardenctl` stays silent and lets `gardenlogin`'s own default apply (currently `auto`).
-
-For an explicit check, the access level is also embedded as a `--access-level=…` argument in the produced kubeconfig's exec plugin section (omitted when `gardenctl` deferred to `gardenlogin`'s default):
-
-```bash
-gardenctl kubeconfig | grep -- '--access-level='
-```
+`gardenctl` can request admin/viewer/auto kubeconfigs from `gardenlogin` per target scope (shoots, seeds), with per-garden defaults and per-invocation overrides via `--access-level` / `--admin` / `--viewer`. See [docs/config/access-level.md](docs/config/access-level.md) for the full reference.
 
 ### Config Path Overwrite
 
--- a/docs/help/gardenctl_config_set-garden.md
+++ b/docs/help/gardenctl_config_set-garden.md
@@ -24,21 +24,26 @@ CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonp
 gardenctl config set-garden $CLUSTER_IDENTITY --kubeconfig $KUBECONFIG
 
 # configure my-garden with a context and patterns
-gardenctl config set-garden my-garden --context garden-context --pattern "^(?:landscape-dev/)?shoot--(?P<project>.+)--(?P<shoot>.+)$" --pattern "https://dashboard\.gardener\.cloud/namespace/(?P<namespace>[^/]+)/shoots/(?P<shoot>[^/]+)
+gardenctl config set-garden my-garden --context garden-context --pattern "^(?:landscape-dev/)?shoot--(?P<project>.+)--(?P<shoot>.+)$" --pattern "https://dashboard\.gardener\.cloud/namespace/(?P<namespace>[^/]+)/shoots/(?P<shoot>[^/]+)"
+
+# configure prd-garden so shoot kubeconfigs default to read-only viewer access (managed seed access stays at admin)
+gardenctl config set-garden prd-garden --default-shoot-access-level viewer
 ```
 
 ### Options
 
 ```
-      --alias string          unique alias of this Garden that can be used instead of the name to target this Garden
-      --context string        override the current-context of the garden cluster kubeconfig
-  -h, --help                  help for set-garden
-      --kubeconfig string     path to kubeconfig file for this Garden cluster
-      --pattern stringArray   define regex match patterns for this garden for custom input formats for targeting.
-                              Use named capturing groups to match target values.
-                              Supported capturing groups: project, namespace, shoot.
-                              Note that if you set this flag it will overwrite the pattern list in the config file.
-                              You may specify any number of extra patterns.
+      --alias string                               unique alias of this Garden that can be used instead of the name to target this Garden
+      --context string                             override the current-context of the garden cluster kubeconfig
+      --default-managed-seed-access-level string   default kubeconfig access level when targeting managed seeds in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
+      --default-shoot-access-level string          default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
+  -h, --help                                       help for set-garden
+      --kubeconfig string                          path to kubeconfig file for this Garden cluster
+      --pattern stringArray                        define regex match patterns for this garden for custom input formats for targeting.
+                                                   Use named capturing groups to match target values.
+                                                   Supported capturing groups: project, namespace, shoot.
+                                                   Note that if you set this flag it will overwrite the pattern list in the config file.
+                                                   You may specify any number of extra patterns.
 ```
 
 ### Options inherited from parent commands
--- a/docs/help/gardenctl_config_set-garden.md
+++ b/docs/help/gardenctl_config_set-garden.md
@@ -26,24 +26,24 @@ gardenctl config set-garden $CLUSTER_IDENTITY --kubeconfig $KUBECONFIG
 # configure my-garden with a context and patterns
 gardenctl config set-garden my-garden --context garden-context --pattern "^(?:landscape-dev/)?shoot--(?P<project>.+)--(?P<shoot>.+)$" --pattern "https://dashboard\.gardener\.cloud/namespace/(?P<namespace>[^/]+)/shoots/(?P<shoot>[^/]+)"
 
-# configure prd-garden so shoot kubeconfigs default to read-only viewer access (managed seed access stays at admin)
+# configure prd-garden so shoot kubeconfigs default to read-only viewer access (seed access stays at admin)
 gardenctl config set-garden prd-garden --default-shoot-access-level viewer
 ```
 
 ### Options
 
 ```
-      --alias string                               unique alias of this Garden that can be used instead of the name to target this Garden
-      --context string                             override the current-context of the garden cluster kubeconfig
-      --default-managed-seed-access-level string   default kubeconfig access level when targeting managed seeds in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
-      --default-shoot-access-level string          default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
-  -h, --help                                       help for set-garden
-      --kubeconfig string                          path to kubeconfig file for this Garden cluster
-      --pattern stringArray                        define regex match patterns for this garden for custom input formats for targeting.
-                                                   Use named capturing groups to match target values.
-                                                   Supported capturing groups: project, namespace, shoot.
-                                                   Note that if you set this flag it will overwrite the pattern list in the config file.
-                                                   You may specify any number of extra patterns.
+      --alias string                        unique alias of this Garden that can be used instead of the name to target this Garden
+      --context string                      override the current-context of the garden cluster kubeconfig
+      --default-seed-access-level string    default kubeconfig access level when targeting seeds in this garden (and shoots that back a managed seed, since they physically are the seed cluster). One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
+      --default-shoot-access-level string   default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
+  -h, --help                                help for set-garden
+      --kubeconfig string                   path to kubeconfig file for this Garden cluster
+      --pattern stringArray                 define regex match patterns for this garden for custom input formats for targeting.
+                                            Use named capturing groups to match target values.
+                                            Supported capturing groups: project, namespace, shoot.
+                                            Note that if you set this flag it will overwrite the pattern list in the config file.
+                                            You may specify any number of extra patterns.
 ```
 
 ### Options inherited from parent commands
--- a/docs/help/gardenctl_config_set-garden.md
+++ b/docs/help/gardenctl_config_set-garden.md
@@ -35,8 +35,8 @@ gardenctl config set-garden prd-garden --default-shoot-access-level viewer
 ```
       --alias string                        unique alias of this Garden that can be used instead of the name to target this Garden
       --context string                      override the current-context of the garden cluster kubeconfig
-      --default-seed-access-level string    default kubeconfig access level when targeting seeds in this garden (and shoots that back a managed seed, since they physically are the seed cluster). One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
-      --default-shoot-access-level string   default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to reset to the built-in default ("admin").
+      --default-seed-access-level string    default kubeconfig access level when targeting seeds in this garden (and shoots that back a managed seed, since they physically are the seed cluster). One of "admin", "viewer", "auto". Pass an empty value to unset and delegate to gardenlogin's default.
+      --default-shoot-access-level string   default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to unset and delegate to gardenlogin's default.
   -h, --help                                help for set-garden
       --kubeconfig string                   path to kubeconfig file for this Garden cluster
       --pattern stringArray                 define regex match patterns for this garden for custom input formats for targeting.
--- a/docs/help/gardenctl_kubectl-env.md
+++ b/docs/help/gardenctl_kubectl-env.md
@@ -13,9 +13,8 @@ For details on how to use the printed shell script, such as applying it temporar
 ### Options
 
 ```
-  -h, --help                             help for kubectl-env
-      --kubeconfig-access-level string   Override default kubeconfig access level for shoots/managed-seeds. One of "admin", "viewer", "auto".
-  -u, --unset                            Generate the script to unset the KUBECONFIG environment variable for 
+  -h, --help    help for kubectl-env
+  -u, --unset   Generate the script to unset the KUBECONFIG environment variable for 
 ```
 
 ### Options inherited from parent commands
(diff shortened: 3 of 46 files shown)


pkg/flags/access_level.go
```diff
@@ -0,0 +1,41 @@
...
+// AddKubeconfigAccessLevelFlag registers the --kubeconfig-access-level flag as a
+// persistent flag on cmd, binding it to value. Shell completion for the flag
+// is registered alongside.
+//
+// This flag is only meaningful for commands that produce a gardenlogin-driven
+// kubeconfig (target, kubeconfig, kubectl-env, ssh, sshpatch). Registering it
+// per-command rather than as a root persistent flag keeps it out of the help
+// output of unrelated commands like `gardenctl config` or `gardenctl version`.
+func AddKubeconfigAccessLevelFlag(cmd *cobra.Command, value *config.KubeconfigAccessLevel) {
+	cmd.PersistentFlags().Var(value, "kubeconfig-access-level",
```
maboehm: To make this a bit less verbose (I suspect this would be the most used part of this), we could use
```
--access=admin|viewer
```
or maybe even `--admin` and `--viewer`

Another idea from the hackathon crowd was to add a `gardenctl break-glass` command to elevate the current target from viewer to admin

pkg/target/manager.go
```diff
@@ -138,16 +154,75 @@ func newGardenClient(name string, config *config.Config, provider internalclient
...
+				configured = garden.DefaultKubeconfigAccessLevel.ManagedSeeds
+			}
+
+			if configured != "" {
+				return configured
+			}
+		}
+	}
+
+	return config.KubeconfigAccessLevelAdmin
```
petersutter: I think we should not default to `admin` here when neither the gardenctl flag nor config specifies an access level.

If gardenctl writes `--access-level=admin` into the generated gardenlogin exec args by default, it overrides gardenlogin’s own defaulting/env-var fallback mechanism. My expectation would be: if gardenctl did not explicitly receive an access level via config or flag, it should omit `--access-level` entirely and let gardenlogin decide.

pkg/cmd/cmd.go
```diff
@@ -120,15 +121,29 @@ Find more information at: https://github.com/gardener/gardenctl-v2/blob/master/R
...
+	targetCmd := cmdtarget.NewCmdTarget(f, ioStreams)
+	kubectlEnvCmd := cmdkubectl.NewCmdKubectlEnv(f, ioStreams)
+	kubeconfigCmd := kubeconfig.NewCmdKubeconfig(f, ioStreams)
+
+	// The --kubeconfig-access-level flag is only meaningful for commands that
+	// produce a gardenlogin-driven kubeconfig.
+	flagsutil.AddKubeconfigAccessLevelFlag(targetCmd, &f.KubeconfigAccessLevel)
+	flagsutil.AddKubeconfigAccessLevelFlag(kubeconfigCmd, &f.KubeconfigAccessLevel)
+	flagsutil.AddKubeconfigAccessLevelFlag(kubectlEnvCmd, &f.KubeconfigAccessLevel)
+	flagsutil.AddKubeconfigAccessLevelFlag(sshCmd, &f.KubeconfigAccessLevel)
```
petersutter: I don’t think this flag should be exposed on `ssh`. The command does not produce a kubeconfig for the user, and it is not obvious which internal client/kubeconfig this flag controls or why a user should influence that as part of SSH setup.

If `ssh` internally needs a shoot/seed client, I would rather keep that as an implementation detail and let the command choose whatever access level is appropriate for its predefined read/list/setup operations. I would not require changing that internal behavior in this PR, but I’d avoid exposing the flag as user-facing API here.

pkg/cmd/cmd.go
```diff
@@ -120,15 +121,29 @@ Find more information at: https://github.com/gardener/gardenctl-v2/blob/master/R
...
+	flagsutil.AddKubeconfigAccessLevelFlag(sshpatchCmd, &f.KubeconfigAccessLevel)
```
petersutter: I think this flag should be dropped from `ssh-patch`. From what I can see, this command uses the garden client / Bastion resources and does not generate or use a gardenlogin-backed shoot/seed kubeconfig where this flag would have an effect. Having it in the help output would therefore be misleading.

pkg/cmd/cmd.go
```diff
@@ -120,15 +121,29 @@ Find more information at: https://github.com/gardener/gardenctl-v2/blob/master/R
...
+	sshpatchCmd := cmdsshpatch.NewCmdSSHPatch(f, ioStreams)
+	targetCmd := cmdtarget.NewCmdTarget(f, ioStreams)
+	kubectlEnvCmd := cmdkubectl.NewCmdKubectlEnv(f, ioStreams)
+	kubeconfigCmd := kubeconfig.NewCmdKubeconfig(f, ioStreams)
+
+	// The --kubeconfig-access-level flag is only meaningful for commands that
+	// produce a gardenlogin-driven kubeconfig.
+	flagsutil.AddKubeconfigAccessLevelFlag(targetCmd, &f.KubeconfigAccessLevel)
+	flagsutil.AddKubeconfigAccessLevelFlag(kubeconfigCmd, &f.KubeconfigAccessLevel)
+	flagsutil.AddKubeconfigAccessLevelFlag(kubectlEnvCmd, &f.KubeconfigAccessLevel)
```
petersutter: For `kubectl-env`, this flag only seems meaningful when symlink mode is disabled and the command writes a fresh kubeconfig via `manager.ClientConfig`.

In symlink mode, `kubectl-env` only points `KUBECONFIG` to the existing session kubeconfig, so `--kubeconfig-access-level` does not affect the generated credentials. Could we either warn in that case or otherwise make the no-op explicit?

pkg/cmd/cmd.go
```diff
@@ -120,15 +121,29 @@ Find more information at: https://github.com/gardener/gardenctl-v2/blob/master/R
...
-	cmd.AddCommand(cmdssh.NewCmdSSH(f, cmdssh.NewSSHOptions(ioStreams)))
-	cmd.AddCommand(cmdsshpatch.NewCmdSSHPatch(f, ioStreams))
-	cmd.AddCommand(cmdtarget.NewCmdTarget(f, ioStreams))
+	sshCmd := cmdssh.NewCmdSSH(f, cmdssh.NewSSHOptions(ioStreams))
+	sshpatchCmd := cmdsshpatch.NewCmdSSHPatch(f, ioStreams)
+	targetCmd := cmdtarget.NewCmdTarget(f, ioStreams)
+	kubectlEnvCmd := cmdkubectl.NewCmdKubectlEnv(f, ioStreams)
+	kubeconfigCmd := kubeconfig.NewCmdKubeconfig(f, ioStreams)
+
+	// The --kubeconfig-access-level flag is only meaningful for commands that
```
petersutter: Could we register this flag in the individual command constructors instead of centrally in the root command?

The helper can still keep the flag definition/completion centralized, but command-local registration makes it clearer which commands actually support the option and avoids coupling command-specific semantics into `NewGardenctlCommand`.

pkg/cmd/config/set_garden.go
```diff
@@ -162,9 +232,57 @@ func (o *setGardenOptions) Run(_ util.Factory) error {
...
 	return nil
 }
 
+// refreshKubeconfigIfTargeted regenerates the active session kubeconfig when
+// the modified garden is the currently-targeted one, so config changes (e.g. a
+// new defaultKubeconfigAccessLevel) take effect for in-flight kubectl calls
+// without the user having to re-target. Soft-fails: the config save already
+// succeeded, so we surface refresh errors as a warning rather than failing the
+// command.
+func (o *setGardenOptions) refreshKubeconfigIfTargeted(f util.Factory) {
```
petersutter: I would avoid refreshing the current session kubeconfig from `config set-garden`.

`config set-garden` changes persistent configuration, while this refresh mutates only the current gardenctl session. Other terminal sessions may still have stale generated kubeconfigs, so this only partially solves the problem and can make behavior harder to reason about.

Could we keep the config command limited to updating config and instead print a general hint that existing session kubeconfigs may need to be regenerated / the target may need to be set again for changes to take effect?

Feels more in line with the KISS principle :)

internal/client/garden/client.go
```diff
@@ -577,7 +586,17 @@ func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clie
...
-		return g.GetShootClientConfig(ctx, "garden", shoot.Name)
+		return g.GetShootClientConfig(ctx, "garden", shoot.Name, accessLevel)
+	}
+
+	// Non-managed seed: kubeconfig comes from a static <name>.login Secret rather
+	// than gardenlogin, so the admin/viewer/auto distinction does not apply.
+	// Reject any explicit non-admin choice rather than silently returning admin -
+	// a silent escalation would defeat the point of asking for viewer.
+	if accessLevel != "" && accessLevel != config.KubeconfigAccessLevelAdmin {
+		return nil, fmt.Errorf("seed %q is not a managed seed; %q kubeconfig access is unavailable (only %q is supported). "+
```
petersutter: I’d rephrase this error slightly. For non-managed seeds, gardenctl does not really know what privileges the static `<seed>.login` kubeconfig grants; it only knows that the gardenlogin access-level mechanism is not available for this path.

So instead of saying that only `admin` is supported, I’d say that access-level selection/requesting `viewer`/`auto` is unavailable for static/non-managed seed kubeconfigs. Failing here is fine when the requested/configured access level cannot be honored, but the wording should avoid implying that gardenctl can determine the actual privilege level of the static kubeconfig.

pkg/target/manager.go
```diff
@@ -138,16 +154,75 @@ func newGardenClient(name string, config *config.Config, provider internalclient
...
+	case t.ControlPlane():
+		return m.resolveAccessLevel(t, AccessScopeManagedSeeds), true
+	case t.ShootName() != "":
+		return m.resolveAccessLevel(t, AccessScopeShoots), true
+	case t.SeedName() != "":
+		return m.resolveAccessLevel(t, AccessScopeManagedSeeds), true
+	default:
+		return "", false
+	}
+}
```
petersutter, line 299: Should this be based on the target path or on the underlying cluster role? For a managed seed backing shoot, `target shoot` and `target seed` can refer to the same cluster but would select different defaults here. I would expect managed seed clusters to use the seed/managed-seed default regardless of how they are targeted, otherwise the access level can depend on targeting preference rather than cluster identity.

pkg/config/config.go
```diff
@@ -57,6 +59,107 @@ type Garden struct {
 	// AccessRestrictions is a list of access restriction definitions
 	// +optional
 	AccessRestrictions []ac.AccessRestriction `json:"accessRestrictions,omitempty"`
+	// DefaultKubeconfigAccessLevel sets the default access level requested per
+	// target scope when no --kubeconfig-access-level flag is provided. Each
+	// sub-field defaults to "admin" when empty.
+	// +optional
+	DefaultKubeconfigAccessLevel *KubeconfigAccessLevels `json:"defaultKubeconfigAccessLevel,omitempty"`
```
petersutter: Naming/API question: would `kubeconfigAccessLevelDefaults` be clearer than `defaultKubeconfigAccessLevel`?

pkg/config/config.go
```diff
@@ -57,6 +59,107 @@ type Garden struct {
...
+// the static seed-login secret, which only supports admin access.
+type KubeconfigAccessLevels struct {
+	// Shoots is the default access level used when targeting a shoot.
+	// +optional
+	Shoots KubeconfigAccessLevel `json:"shoots,omitempty"`
+	// ManagedSeeds is the default access level used when targeting a managed seed
+	// (or a shoot's control plane on a managed seed).
+	// +optional
+	ManagedSeeds KubeconfigAccessLevel `json:"managedSeeds,omitempty"`
+}
```
petersutter, line 83: Not asking to implement this in this PR, but I think it is worth briefly considering the extensibility of this config shape.

This feature is mostly operator-oriented right now, but similar defaults could later be useful for more granular cases, e.g. project/shoot-specific defaults or labels/annotations marking sensitive shoots. I would not want to over-engineer the structure now, but we should avoid choosing a schema that would be hard to extend compatibly later.

pkg/cmd/kubeconfig/kubeconfig.go
```diff
@@ -124,18 +125,27 @@ func (o *options) Complete(f util.Factory, _ *cobra.Command, _ []string) error {
...
 	o.RawConfig = &rawConfig
 
+	// Always surface the kubeconfig access level so the user can be confident
+	// which credentials they got. Goes to stderr so it doesn't pollute the
+	// kubeconfig YAML on stdout (which users typically pipe to a file or kubectl).
+	// Skipped for targets that don't produce a gardenlogin kubeconfig (e.g.
+	// garden- or project-only targets) where the notion does not apply.
+	if level, ok := manager.EffectiveAccessLevel(currentTarget); ok && level != "" {
+		fmt.Fprintf(o.IOStreams.ErrOut, "Kubeconfig access level: %s\n", level)
+	}
```
petersutter: I’d only print this when gardenctl explicitly requested an access level via flag/config.

If no access level was set in gardenctl, and we omit `--access-level` so gardenlogin can apply its own default/env fallback, then gardenctl should probably not print an access level either because it did not actually decide one.

Also I’d keep the wording short, e.g. `Access level: viewer`. The command context already makes it clear that this refers to the kubeconfig.

pkg/cmd/target/target.go
```diff
@@ -228,6 +228,14 @@ func (o *TargetOptions) Run(f util.Factory) error {
 		} else if o.Kind != "" {
 			fmt.Fprintf(o.IOStreams.Out, "Successfully targeted %s %q\n", o.Kind, o.TargetName)
 		}
+
+		// Always surface the kubeconfig access level so the user can be confident
+		// which credentials they got. Skipped only for targets that don't produce
+		// a gardenlogin kubeconfig (garden- or project-only targets), where the
+		// notion does not apply.
+		if level, ok := manager.EffectiveAccessLevel(currentTarget); ok && level != "" {
+			fmt.Fprintf(o.IOStreams.Out, "Kubeconfig access level: %s\n", level)
```
petersutter: Minor UX suggestion: for `target`, I’d keep the existing concise success output and append this only when gardenctl explicitly requested an access level, e.g.

`Successfully targeted shoot "prod" (access level: viewer)`

This avoids adding another line while still making the chosen access level visible.

petersutter (commented):
Thanks for working on this, I like the feature direction 🚀 . I added a few inline comments for the concrete places where I think the behavior/API should be clarified.

My main themes are:

- gardenctl should not emit `--access-level` unless the level was explicitly configured or passed as a gardenctl flag, so gardenlogin’s own default/env fallback keeps working
- the flag should only be exposed on commands where it has a clear user-facing effect
- the seed/managed-seed semantics should be clarified so the selected default does not unexpectedly depend on targeting path
- config commands should avoid mutating only the current session kubeconfig as side effect
- access-level output should be concise and only shown when gardenctl actually requested a specific level

docs/help/gardenctl_kubectl-env.md
```diff
@@ -13,8 +13,9 @@ For details on how to use the printed shell script, such as applying it temporar
 ### Options
 
 ```
-  -h, --help    help for kubectl-env
-  -u, --unset   Generate the script to unset the KUBECONFIG environment variable for 
+  -h, --help                             help for kubectl-env
+      --kubeconfig-access-level string   Override default kubeconfig access level for shoots/managed-seeds. One of "admin", "viewer", "auto".
+  -u, --unset                            Generate the script to unset the KUBECONFIG environment variable for
```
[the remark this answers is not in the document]
jamand: Pre-existing, also marked as outdated.

pkg/flags/access_level.go
jamand: I did two changes here: the verbose flag is now called `--access-level` (consistent with `gardenlogin`) and there is now short forms `--viewer` and `--admin` (added in 6d5c011). For the `gardenctl break-glass` I did not add that here, as I feel I already added a few UX/usability things outside the core feature that made the PR grow and repeating the last target command with the flag should be easy enough. In case it shows to be annoying I'm happy to add it though :)

pkg/target/manager.go
jamand: Good point, changed to not passing an argument to `gardenlogin` in case no flag or config is passed. (25d6408)

pkg/cmd/cmd.go
jamand: Ah, yes. I got confused by the fact that `manager.ShootClient`/`SeedClient` is also called by the `ssh` command but that's only an implementation detail and it does not make sense to let the operator configure something here! (f19c553)

pkg/cmd/cmd.go
jamand: Similar to `ssh`: (f19c553)

pkg/cmd/cmd.go
jamand: Fixed in f350c2e by dropping the flag entirely, is that fine by you or should we do more special handling for the non-symlink mode?

pkg/cmd/cmd.go
jamand: Done in 8f2f05e (using shared code for the flag)

pkg/cmd/config/set_garden.go
jamand: I went with the hint/warning route, I think when I noticed that I went a bit into overengineering-mode :)
(88b13c4)

internal/client/garden/client.go
jamand: Rephrased in fbec3f9

pkg/target/manager.go
jamand: I addressed this in c43f44b. This is a very good point I honestly did not consider.

I added `Client.IsManagedSeed` (using the same `spec.shoot.name` field selector as upstream Gardener's `GetManagedSeedWithReader`) and consult it from `ClientConfig` and `EffectiveAccessLevel` for shoot targets. A managed-seed-backing shoot now routes to the `managedSeeds` scope regardless of whether the user reaches it via `target shoot` or `target seed`, so it's again about what is accessed and not how it's accessed.

pkg/config/config.go
jamand: Sounds good (6024494)

pkg/cmd/kubeconfig/kubeconfig.go
jamand: Applied in a68c3a8 (wording + only print on explicit setting)

pkg/cmd/target/target.go
jamand: Done together with the last comment (on explicit setting): a68c3a8

jamand:
I think the CodeRabbit bot might have overwritten edits I did in the PR description, my first edit is missing, my second one is there :/

internal/client/garden/client.go
```diff
@@ -577,7 +605,21 @@ func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clie
 			},
 			"seed", name)
 
-		return g.GetShootClientConfig(ctx, "garden", shoot.Name)
+		return g.GetShootClientConfig(ctx, "garden", shoot.Name, accessLevel)
+	}
+
+	// Non-managed seed: kubeconfig comes from a static <name>.login Secret, so
```
maboehm, line 639: One small UX nit I noticed here: When I target a non-managed seed while having `kubeconfigAccessLevelDefaults.managedSeeds: admin` it still logged `(access level: admin)`, which is a bit misleading, because we cannot tell what the access level of the `seed.login` kubeconfig is.

pkg/config/config.go
jamand: Agreed that this is worth keeping in mind to avoid breaking changes in the future. From my understanding, I would think that we could have the per-garden defaults I introduced here + a rules block with an arbitrary number of (first-match) blocks that match projects or labels, which are then used to specify a deviation from the default values.

jamand:
Hello @petersutter @maboehm ! Thanks for the reviews again. I addressed the comments. For a few questions a posted an answer which I hope resolves it, in that case feel free to close it. Otherwise I'm happy to provide more input.

internal/client/garden/client.go
```diff
@@ -381,6 +400,36 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
...
+	}
+
+	for i := range list.Items {
+		if list.Items[i].Name == seedName {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
```
petersutter, line 438: Per docs (https://gardener.cloud/docs/gardener/managed_seed/), ManagedSeeds are restricted to the garden namespace. Cluster-wide List+filter is unnecessary - use Get directly. See `gardener/gardener pkg/utils/kubernetes/managedseed.go` `GetManagedSeedByName` for the established pattern: https://github.com/gardener/gardener/blob/576cbc70ec5fd3a7f46e7eb604191854a00fb74e/pkg/utils/kubernetes/managedseed.go#L37-L46
Also note that regular users typically do not have the permission to get/list managedSeed resources (as mentioned in a another comment)

internal/client/garden/client.go
```diff
@@ -68,8 +71,13 @@ type Client interface {
...
 	// ListSeeds returns all Gardener seed resources
 	ListSeeds(ctx context.Context, opts ...client.ListOption) (*gardencorev1beta1.SeedList, error)
-	// GetSeedClientConfig returns the client config for a seed
-	GetSeedClientConfig(ctx context.Context, name string) (clientcmd.ClientConfig, error)
+	// GetSeedClientConfig returns the client config for a seed.
+	// For managed seeds, accessLevel is honored (the call is delegated to
+	// GetShootClientConfig). For non-managed seeds the kubeconfig comes from a
+	// static <name>.login Secret which only supports admin access - a non-admin
+	// accessLevel returns an error rather than silently producing an admin
+	// kubeconfig. Pass an empty value or "admin" for non-managed seeds.
```
petersutter:
```suggestion
	// GetSeedClientConfig returns the client config for a seed.
	//
	// For managed seeds, accessLevel is honored (the call is delegated to
	// GetShootClientConfig).
	//
	// For non-managed seeds the kubeconfig comes from a static <name>.login
	// Secret; an explicit "viewer" accessLevel returns an error since gardenctl
	// cannot guarantee the privileges of a static kubeconfig. Any other value
	// (empty, "admin", "auto") returns the static kubeconfig as-is.
```

internal/client/garden/client.go
```diff
@@ -100,6 +110,15 @@ type Client interface {
 	GetConfigMap(ctx context.Context, namespace, name string) (*corev1.ConfigMap, error)
 	// GetShootOfManagedSeed returns shoot of seed using ManagedSeed resource. An error is returned if it is not a managed seed or the referenced shoot is nil
 	GetShootOfManagedSeed(ctx context.Context, name string) (*seedmanagementv1alpha1.Shoot, error)
+	// IsManagedSeed reports whether the shoot identified by (shootNamespace, shootName)
+	// is referenced by a ManagedSeed resource. Uses the spec.shoot.name field
+	// selector rather than relying on the (conventional but not guaranteed)
+	// equality of ManagedSeed and shoot names.
```
petersutter:
```suggestion
	// is referenced by a ManagedSeed resource.
```

internal/client/garden/client.go
```diff
@@ -100,6 +110,15 @@ type Client interface {
...
+	IsManagedSeed(ctx context.Context, shootNamespace, shootName string) (bool, error)
+	// IsManagedSeedByName reports whether a seed of the given name is a managed
+	// seed (i.e. produced by a ManagedSeed resource). Lists cluster-wide so it
+	// works regardless of which namespace the ManagedSeed lives in.
```
petersutter: > Lists cluster-wide

as written below, we should limit it to the "garden" namespace

internal/client/garden/client.go
```diff
@@ -381,6 +400,36 @@ func (g *clientImpl) GetConfigMap(ctx context.Context, namespace, name string) (
...
+
+	if err := g.c.List(ctx, list,
+		client.InNamespace(shootNamespace),
+		client.MatchingFields{seedmanagement.ManagedSeedShootName: shootName},
+	); err != nil {
+		return false, err
+	}
+
+	return len(list.Items) > 0, nil
+}
```
petersutter, line 423: early-return false when shootNamespace != "garden" to skip the API call entirely. ManagedSeeds restricted to `garden` (see other comment below), so the list request is wasted for any other namespace. In addition it currently breaks the target command for regular users that do not have the permission to list ManagedSeeds, which results in
```
Error: managedseeds.seedmanagement.gardener.cloud is forbidden: User "me@example.com" cannot list resource "managedseeds" in API group "seedmanagement.gardener.cloud" in the namespace "garden-myproject"
```

internal/client/garden/client.go
```diff
@@ -577,7 +626,21 @@ func (g *clientImpl) GetSeedClientConfig(ctx context.Context, name string) (clie
...
+	// reject only an explicit "viewer" request - the user asked for read-only
+	// and we cannot deliver it, so silently returning the static kubeconfig
+	// would be a privilege surprise. "auto" semantically means "try admin,
+	// fall back to viewer", so falling through here matches its documented
+	// behavior. Note that gardenctl cannot determine the actual privileges
+	// the static kubeconfig grants - they depend on the cluster's RBAC.
+	if accessLevel == config.KubeconfigAccessLevelViewer {
+		return nil, fmt.Errorf("seed %q is not a managed seed; gardenlogin's access-level mechanism is unavailable for static seed-login kubeconfigs. "+
+			"Re-run without %q (or with --access-level=%s) to use the static kubeconfig as-is",
+			name, accessLevel, config.KubeconfigAccessLevelAdmin)
```
petersutter:
```suggestion
			name, accessLevel, config.KubeconfigAccessLevelAuto)
```
recommending auto matches user intent without overclaiming admin

pkg/cmd/config/set_garden.go
```diff
@@ -72,16 +84,57 @@ type setGardenOptions struct {
 	// Supported capturing groups: project, namespace, shoot
 	// +optional
 	Patterns []string
+	// DefaultShootAccessLevelFlag sets kubeconfigAccessLevelDefaults.shoots in the
+	// stored Garden config (admin | viewer | auto).
+	// +optional
+	DefaultShootAccessLevelFlag accessLevelFlag
+	// DefaultManagedSeedAccessLevelFlag sets kubeconfigAccessLevelDefaults.managedSeeds
+	// in the stored Garden config (admin | viewer | auto).
```
petersutter:
```suggestion
	// in the stored Garden config (admin | viewer | auto  | "" to reset to built-in default).
```

pkg/cmd/config/set_garden.go
```diff
@@ -72,16 +84,57 @@ type setGardenOptions struct {
 	// Supported capturing groups: project, namespace, shoot
 	// +optional
 	Patterns []string
+	// DefaultShootAccessLevelFlag sets kubeconfigAccessLevelDefaults.shoots in the
+	// stored Garden config (admin | viewer | auto).
```
petersutter:
```suggestion
	// stored Garden config (admin | viewer | auto | "" to reset to built-in default).
```

pkg/cmd/config/set_garden.go
```diff
@@ -72,16 +84,57 @@ type setGardenOptions struct {
...
+}
+
+// accessLevelFlag is a pflag.Value for config.KubeconfigAccessLevel that also
+// tracks whether the flag was explicitly provided on the command line. This
+// lets set-garden distinguish "user wants to set to X" from "user did not
+// touch this field" - important when updating an existing Garden entry.
+type accessLevelFlag struct {
+	provided bool
+	value    config.KubeconfigAccessLevel
 }
```
petersutter, line 102: How about embedding flag.StringFlag here instead of reimplementing the provided-tracking + accessor methods? It already provides `Provided()`/`Value()`/`String()`/`Type()` with the exact same semantics, so we only need to override Set for validation and add a typed AccessLevel() accessor. Keeps this consistent with the other flags in this command (Alias, KubeconfigFlag, ContextFlag) which all use flag.StringFlag.

```diff
diff --git a/pkg/cmd/config/set_garden.go b/pkg/cmd/config/set_garden.go
index 03dc15b..7566c92 100644
--- a/pkg/cmd/config/set_garden.go
+++ b/pkg/cmd/config/set_garden.go
@@ -94,33 +94,26 @@ type setGardenOptions struct {
 	DefaultManagedSeedAccessLevelFlag accessLevelFlag
 }
 
-// accessLevelFlag is a pflag.Value for config.KubeconfigAccessLevel that also
-// tracks whether the flag was explicitly provided on the command line. This
-// lets set-garden distinguish "user wants to set to X" from "user did not
-// touch this field" - important when updating an existing Garden entry.
+// accessLevelFlag wraps flag.StringFlag to validate config.KubeconfigAccessLevel
+// values at parse time while inheriting the standard provided-tracking semantics
+// used by the other flags in this command.
 type accessLevelFlag struct {
-	provided bool
-	value    config.KubeconfigAccessLevel
+	flag.StringFlag
 }
 
 var _ pflag.Value = (*accessLevelFlag)(nil)
 
 func (f *accessLevelFlag) Set(value string) error {
-	candidate := config.KubeconfigAccessLevel(value)
-	if err := candidate.Validate(); err != nil {
+	if err := config.KubeconfigAccessLevel(value).Validate(); err != nil {
 		return err
 	}
 
-	f.value = candidate
-	f.provided = true
-
-	return nil
+	return f.StringFlag.Set(value)
 }
 
-func (f *accessLevelFlag) Type() string                        { return "string" }
-func (f *accessLevelFlag) String() string                      { return string(f.value) }
-func (f *accessLevelFlag) Provided() bool                      { return f.provided }
-func (f *accessLevelFlag) Value() config.KubeconfigAccessLevel { return f.value }
+func (f *accessLevelFlag) AccessLevel() config.KubeconfigAccessLevel {
+	return config.KubeconfigAccessLevel(f.Value())
+}
 
 // Complete adapts from the command line args to the data required.
 func (o *setGardenOptions) Complete(f util.Factory, _ *cobra.Command, args []string) error {
@@ -248,11 +241,11 @@ func (o *setGardenOptions) applyAccessLevelFlags(garden *config.Garden) {
 	}
 
 	if o.DefaultShootAccessLevelFlag.Provided() {
-		garden.KubeconfigAccessLevelDefaults.Shoots = o.DefaultShootAccessLevelFlag.Value()
+		garden.KubeconfigAccessLevelDefaults.Shoots = o.DefaultShootAccessLevelFlag.AccessLevel()
 	}
 
 	if o.DefaultManagedSeedAccessLevelFlag.Provided() {
-		garden.KubeconfigAccessLevelDefaults.ManagedSeeds = o.DefaultManagedSeedAccessLevelFlag.Value()
+		garden.KubeconfigAccessLevelDefaults.ManagedSeeds = o.DefaultManagedSeedAccessLevelFlag.AccessLevel()
 	}
 
 	if garden.KubeconfigAccessLevelDefaults.Shoots == "" && garden.KubeconfigAccessLevelDefaults.ManagedSeeds == "" {
```

pkg/cmd/config/set_garden.go
```diff
@@ -119,6 +172,12 @@ Use named capturing groups to match target values.
 Supported capturing groups: project, namespace, shoot.
 Note that if you set this flag it will overwrite the pattern list in the config file.
 You may specify any number of extra patterns.`)
+	flags.Var(&o.DefaultShootAccessLevelFlag, FlagDefaultShootAccessLevel,
+		fmt.Sprintf(`default kubeconfig access level when targeting shoots in this garden. One of %q, %q, %q. Pass an empty value to reset to the built-in default (%q).`,
+			config.KubeconfigAccessLevelAdmin, config.KubeconfigAccessLevelViewer, config.KubeconfigAccessLevelAuto, config.KubeconfigAccessLevelAdmin))
+	flags.Var(&o.DefaultManagedSeedAccessLevelFlag, FlagDefaultManagedSeedAccessLevel,
+		fmt.Sprintf(`default kubeconfig access level when targeting managed seeds in this garden. One of %q, %q, %q. Pass an empty value to reset to the built-in default (%q).`,
+			config.KubeconfigAccessLevelAdmin, config.KubeconfigAccessLevelViewer, config.KubeconfigAccessLevelAuto, config.KubeconfigAccessLevelAdmin))
```
petersutter: Help text claims built-in default is admin, but unsetting actually returns empty from resolveAccessLevel (manager.go:306) — gardenctl then omits --access-level and gardenlogin's own default applies. Reword to avoid pinning a specific value here.

```suggestion
		fmt.Sprintf(`default kubeconfig access level when targeting shoots in this garden. One of %q, %q, %q. Pass an empty value to unset and delegate to gardenlogin's default.`,
			config.KubeconfigAccessLevelAdmin, config.KubeconfigAccessLevelViewer, config.KubeconfigAccessLevelAuto))
	flags.Var(&o.DefaultManagedSeedAccessLevelFlag, FlagDefaultManagedSeedAccessLevel,
		fmt.Sprintf(`default kubeconfig access level when targeting managed seeds in this garden. One of %q, %q, %q. Pass an empty value to unset and delegate to gardenlogin's default.`,
			config.KubeconfigAccessLevelAdmin, config.KubeconfigAccessLevelViewer, config.KubeconfigAccessLevelAuto))
```

pkg/cmd/config/set_garden.go
```diff
@@ -162,9 +227,39 @@ func (o *setGardenOptions) Run(_ util.Factory) error {
 
 	fmt.Fprintf(o.IOStreams.Out, "Successfully configured garden %q\n", o.Name)
 
+	if o.DefaultShootAccessLevelFlag.Provided() || o.DefaultManagedSeedAccessLevelFlag.Provided() {
+		fmt.Fprintf(o.IOStreams.ErrOut,
+			"Note: existing session kubeconfigs are not regenerated. Run `gardenctl target` again for the new access level to take effect.\n")
```
petersutter:
```suggestion
			"\nNote: existing session kubeconfigs are not regenerated. Run `gardenctl target` again for the new access level to take effect.\n")
```
nit. imo this reads better. This then looks like:

```sh
$ gardenctl config set-garden prd-garden --default-shoot-access-level viewer
Successfully configured garden "prd-garden"

Note: existing session kubeconfigs are not regenerated. Run `gardenctl target` again for the new access level to take effect.
```

pkg/target/manager.go
```diff
@@ -432,6 +590,14 @@ func (m *managerImpl) updateTarget(ctx context.Context, target Target) error {
 
 func (m *managerImpl) ClientConfig(ctx context.Context, t Target) (clientcmd.ClientConfig, error) {
 	if t.ControlPlane() {
+		// A shoot's control plane runs as pods in its seed; this branch returns the
+		// seed kubeconfig (namespaced to the shoot CP). The seed under a shoot CP is
+		// typically a managed seed, so the managed-seeds scope is the right default.
+		// If the seed turns out to be non-managed, GetSeedClientConfig rejects any
+		// non-admin level - the user can override per invocation with
+		// --access-level=admin.
+		accessLevel := m.resolveAccessLevel(t, AccessScopeManagedSeeds)
```
petersutter: I think the intended behavior here needs to be made more explicit.

As I understand it, the PR wants to protect users who request `viewer` access from accidentally receiving a static seed kubeconfig that may have broader privileges. That makes sense: for non-managed seeds, gardenctl cannot request admin/viewer credentials via gardenlogin, and it also cannot know what permissions the static seed-login kubeconfig actually grants.

My concern is the naming/semantics of the config default. 
```yaml
kubeconfigAccessLevelDefaults:
  managedSeeds: viewer
```
reads to me as “use viewer when targeting managed seeds”. But this value is passed into `GetSeedClientConfig` before we know whether the resolved seed is managed. As a result, this config can also make non-managed seed targeting fail, even though the key sounds scoped to managed seeds only.

Maybe the broader semantic is actually “seed kubeconfig default/safety policy” rather than “managed seed default”. For example, a `seeds` default would be easier to document: on managed seeds, `viewer` can be requested via gardenlogin; on non-managed seeds, `viewer` intentionally fails because gardenctl only has the static kubeconfig and cannot guarantee it is read-only.

Same concern applies to the direct Seed branch (`if t.SeedName() != "" { ... }`) below.

pkg/cmd/cmd.go
petersutter: I think dropping the flag entirely is probably not quite right.

By default, symlink mode is enabled, and in that mode `kubectl-env` only emits the script that points `KUBECONFIG` to the stable session kubeconfig. So in that case the access-level flag would indeed be a no-op.

However, with `linkKubeconfig: false`, `kubectl-env` is still the command that actually creates/downloads the kubeconfig for the current target. For that mode, having the access-level flag on `kubectl-env` is semantically correct.

That said, I’m also not sure whether anyone still actively uses gardenctl with symlink mode disabled. This was the initial behavior before the symlink mode was introduced, and we could consider deprecating/removing the non-symlink mode separately in the future.

For this PR, I’d prefer to restore the flag for `kubectl-env`, but make the no-op explicit when symlink mode is enabled. For example, we could warn only if the flag is explicitly set while `linkKubeconfig` is true, or mention in the help text that the flag only has an effect with `linkKubeconfig: false`.

pkg/config/config.go
petersutter: sounds good, thanks

pkg/cmd/kubeconfig/kubeconfig.go
```diff
@@ -19,12 +20,14 @@ import (
...
+	"github.com/gardener/gardenctl-v2/pkg/config"
 	"github.com/gardener/gardenctl-v2/pkg/flags"
 	"github.com/gardener/gardenctl-v2/pkg/target"
 )
 
-// NewCmdKubeconfig returns a new kubeconfig command.
-func NewCmdKubeconfig(f util.Factory, ioStreams util.IOStreams) *cobra.Command {
+// NewCmdKubeconfig returns a new kubeconfig command. accessLevel is bound to
+// the --access-level flag.
+func NewCmdKubeconfig(f util.Factory, ioStreams util.IOStreams, accessLevel *config.KubeconfigAccessLevel) *cobra.Command {
```
petersutter, line 30: The `accessLevel *config.KubeconfigAccessLevel` parameter here (and on `NewCmdTarget`) is a symptom of `KubeconfigAccessLevel` living on `FactoryImpl`. I think that's the wrong home for it:

- It's a per-command flag, but stored as process-level bootstrap state alongside `GardenHomeDirectory`, `ConfigFile`, etc.
- The factory becomes a mailbox between cobra and `managerImpl.flagAccessLevel` (snapshotted in `NewManager` at `manager.go:165`), which is why we need the pointer threading and cobra has to write back into the same memory the manager later reads.

If we drop the field from `FactoryImpl` and have each command that wants the flag own it on its own options struct (passing it explicitly to manager methods that need it), `NewCmdKubeconfig(f, ioStreams)` and `NewCmdTarget(f, ioStreams)` match the rest of the repo, and the flag's storage scope matches its actual scope.

Happy to defer this to a follow-up PR if you'd rather not expand the diff here.

pkg/cmd/target/target.go
```diff
@@ -223,10 +227,37 @@ func (o *TargetOptions) Run(f util.Factory) error {
...
+		// sees it instead of falling through to "no scope".
+		levelTarget := currentTarget
+		switch o.Kind {
+		case TargetKindShoot:
+			levelTarget = levelTarget.WithShootName(o.TargetName)
+		case TargetKindSeed:
+			levelTarget = levelTarget.WithSeedName(o.TargetName)
+		case TargetKindControlPlane:
+			levelTarget = levelTarget.WithControlPlane(true)
+		}
```
petersutter: Agree it's a smell (predates this PR)

FYI I Filed #744 for the root-cause fix. Once fixed we can drop this part.

pkg/cmd/target/target.go
```diff
@@ -223,10 +227,37 @@ func (o *TargetOptions) Run(f util.Factory) error {
...
+
+		var levelSuffix string
+
+		level, ok, lvlErr := manager.EffectiveAccessLevel(ctx, levelTarget)
```
petersutter, line 249: `EffectiveAccessLevel` here recomputes the scope that `ClientConfig` already resolved during `updateClientConfigSymlink`. The concern is correctness and ownership: two code paths independently deciding the access level can drift, and neither path clearly owns the decision. If they disagree, the displayed level won't match what's actually in the kubeconfig.

A single source of truth would also let us drop the `levelTarget` workaround above (re-applying the just-targeted leaf so `EffectiveAccessLevel` sees it), which I referenced in #744.

pkg/flags/access_level.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (b *boolAccessLevel) Set(v string) error {
+	parsed, err := strconv.ParseBool(v)
+	if err != nil {
+		return fmt.Errorf("invalid boolean value %q", v)
+	}
+
+	if !parsed {
+		return fmt.Errorf("does not accept false; omit the flag instead, or use --access-level=... to pick a different level")
+	}
```
petersutter, line 75: why are we so strict here? Shouldn't we treat it as if it were omitted instead?

README.md
```diff
@@ -90,12 +90,77 @@ gardens:
...
+# kubeconfigAccessLevelDefaults: # Default access level requested per target scope
+#   shoots: viewer              # admin (default) | viewer | auto
+#   managedSeeds: admin         # admin (default) | viewer | auto
 ```
 
 > [!NOTE]
 > - To use the kubeconfigs for **shoot clusters** provided by `gardenctl`, you need to have [gardenlogin](https://github.com/gardener/gardenlogin) installed as a `kubectl` auth plugin.
 > - If your **garden cluster** kubeconfig uses OIDC authentication, ensure that you have the [kubelogin](https://github.com/int128/kubelogin) `kubectl` auth plugin installed.
 
+### Kubeconfig Access Level
```
petersutter, line 102: This is useful documentation, but it’s quite detailed for the README. Could we move the full explanation to a dedicated file under `/docs` and keep only a brief README mention/link near the config example?

docs/help/gardenctl_kubectl-env.md
```diff
@@ -9,12 +9,17 @@ Generate a script that points KUBECONFIG to the currently targeted shoot, seed,
...
 
 ```
-  -h, --help    help for kubectl-env
-  -u, --unset   Generate the script to unset the KUBECONFIG environment variable for 
+      --access-level string   Override the default kubeconfig access level when targeting shoots or seeds. One of "admin", "viewer", "auto".
+      --admin                 shorthand for --access-level=admin
+  -h, --help                  help for kubectl-env
+  -u, --unset                 Generate the script to unset the KUBECONFIG environment variable for 
+      --viewer                shorthand for --access-level=viewer
 ```
```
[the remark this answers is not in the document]
jamand: The other help docs do not do that either, skipped.

docs/help/gardenctl_config_set-garden.md
```diff
@@ -24,21 +24,26 @@ CLUSTER_IDENTITY=$(kubectl -n kube-system get configmap cluster-identity -ojsonp
...
+      --default-shoot-access-level string   default kubeconfig access level when targeting shoots in this garden. One of "admin", "viewer", "auto". Pass an empty value to unset and delegate to gardenlogin's default.
+  -h, --help                                help for set-garden
+      --kubeconfig string                   path to kubeconfig file for this Garden cluster
+      --pattern stringArray                 define regex match patterns for this garden for custom input formats for targeting.
+                                            Use named capturing groups to match target values.
+                                            Supported capturing groups: project, namespace, shoot.
+                                            Note that if you set this flag it will overwrite the pattern list in the config file.
+                                            You may specify any number of extra patterns.
 ```
```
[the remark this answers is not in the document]
jamand: The other help docs do not do that either, skipped.

pkg/cmd/cmd.go
jamand: Understood. I never used `gardenctl` with symlink mode disabled, so I was a bit unsure about that part. It's now handled in df51e9029e96f6eaaa2820d940c498a471c7c5d4.

internal/client/garden/client.go
jamand: Good point. Somehow I thought that was a historic restriction that is lifted now but obviously it's not. Adapted in d5ba24820f8ef8c5e24522294703c4eb06cf2f1c (the bigger of the added commits). Also there is now a check at the beginning of `IsManagedSeed` that checks if the shoot is in the garden namespace, if it's not it cannot be backed by a `ManagedSeed` (and does not need to do a GET thus not needing the permissions).

internal/client/garden/client.go
jamand: See above, adapted in https://github.com/gardener/gardenctl-v2/commit/d5ba24820f8ef8c5e24522294703c4eb06cf2f1c.

pkg/cmd/config/set_garden.go
jamand: Done here: https://github.com/gardener/gardenctl-v2/pull/735/changes/1d73589cc9eb392f7c5cf3f1e69e260ec41bf054

README.md
jamand: Good point, moved the docs + added a link: https://github.com/gardener/gardenctl-v2/pull/735/changes/ddc651562be890438dd3f119b96cdf8a297adc69

pkg/flags/access_level.go
jamand: From my understanding there is no easy way to get that working in Cobra without adding a lot of manual logic (`MarkFlagsMutuallyExclusive` would flag `--viewer=false --admin` because both flags are provided even if one is false). I added a new comment about it in the line https://github.com/gardener/gardenctl-v2/pull/735/changes/1d73589cc9eb392f7c5cf3f1e69e260ec41bf054#diff-43351af59975108575aba6ca7ec633aa301c30458931ef359f866180cb9e09cbR69
Using `--admin`, `--viewer`, `--access-level auto` should be switches to turn it on, doing things like `--viewer=false --admin` could theoretically be possible but don't add much value, so the idea is to prompt the user to just do `--admin` and to disallow the false options.

pkg/cmd/target/target.go
jamand: Nice, I added a TODO + reference to the issue.

pkg/cmd/kubeconfig/kubeconfig.go
jamand: I would do that in a follow-up PR, as is touches a series of functions/files making the review heavier. Here's already the issue https://github.com/gardener/gardenctl-v2/issues/747 and a dummy PR https://github.com/jamand/gardenctl-v2/pull/1 in my fork (which I would then re-create in this repo).

pkg/target/manager.go
jamand: The renaming makes the semantics a lot clearer I think, I went with it in d5ba24820f8ef8c5e24522294703c4eb06cf2f1c. And the case of "`viewer` cannot be guaranteed for non-managed seeds" becomes then transparent to the user when they try to do it (error message).

pkg/cmd/target/target.go
jamand: I did an improvement to reduce the drift risk in https://github.com/gardener/gardenctl-v2/pull/735/changes/b2221e5cfb624b6af574210f8556972f4b45f751, so both `EffectiveAccessLevel` and `ClientConfig` return the same level for the same target. To my understanding I'll be able to do a follow-up and improvement after #744 is fixed to really have a single source of truth, only in that case I can remove the `levelTarget` workaround and cache the level to only have a single level resolution call. I can gladly create a follow-up issue for that.

jamand:
@petersutter I addressed the open suggestions + referred to an open issue + linked a potential implementation for the refactor (move the field from `FactoryImpl` to a more reasonable place) as per your suggestion (to not bloat the diff, as it's small but scattered changes. See https://github.com/gardener/gardenctl-v2/pull/735#discussion_r3309251716

pkg/cmd/target/target.go
petersutter: > I can gladly create a follow-up issue for that.

yes that would be good, thanks

petersutter (commented):
/lgtm
/approve

thanks!

PR #766, opened 2026-06-22
refactor(target): unify target setters and overlays

### LINKED ISSUE #744: CurrentTarget() is not idempotent
## Summary

`Manager.CurrentTarget()` does not reliably return the most recently persisted target when CLI target flags (notably `--garden`) are set. Calling `Write(t)` followed by `CurrentTarget()` can return a target with deeper levels (project/seed/shoot/control-plane) wiped, even though they were just written to disk.

This violates the basic expectation that a getter named `CurrentTarget()` returns "what is currently targeted".

## Root cause

`dynamicTargetProvider.Read()` re-runs `merge(diskTarget, targetFlags)` on every call. The `merge` function applies a "move up" rule:

https://github.com/gardener/gardenctl-v2/blob/master/pkg/target/target_provider.go#L163-L169

```go
// Setting a garden resets all deeper targeting levels, allowing
// the user to "move up". For example, when they have targeted a shoot,
// simply specifying "--garden mygarden" should target the garden, not
// the same shoot within the garden named mygarden.
if tf.GardenName() != "" {
    newTarget = newTarget.WithGardenName(tf.GardenName()).
        WithProjectName("").WithSeedName("").WithShootName("")
}
```

This rule is correct for the *first* read of an invocation (where the flags express user intent). It is wrong for *subsequent* reads within the same invocation, after `TargetXXX` has already persisted the resolved target to disk — the same flags fire again and wipe the leaves the command itself just wrote.

## Reproduction

```sh
gardenctl target shoot my-shoot --garden my-garden --project my-project
```

1. `target.Run` calls `manager.TargetShoot("my-shoot")`.
2. `TargetShoot` calls `CurrentTarget()` (merge applied — fine, this is the first read), builds `{garden, project, shoot}`, writes to disk.
3. `target.Run` calls `manager.CurrentTarget()` again for the success message.
4. `merge` runs again on fresh disk state, sees `--garden` is set, wipes shoot + project. Returned target: `{garden=my-garden, project="", shoot=""}`.
5. Display code has to manually re-apply `o.TargetName` to recover what was just persisted.

The display-path workaround was introduced in PR #735 (not yet merged) as a local fix because `EffectiveAccessLevel` otherwise saw "no scope". On the PR branch it lives at `pkg/cmd/target/target.go` around the `levelTarget := currentTarget` block:

```go
// dynamicTargetProvider.merge() wipes deeper target levels whenever
// --garden is set on the CLI, even immediately after TargetXXX persisted
// a shoot/seed. Re-apply the just-targeted leaf so EffectiveAccessLevel
// sees it instead of falling through to "no scope".
levelTarget := currentTarget
switch o.Kind {
case TargetKindShoot:
    levelTarget = levelTarget.WithShootName(o.TargetName)
case TargetKindSeed:
    levelTarget = levelTarget.WithSeedName(o.TargetName)
case TargetKindControlPlane:
    levelTarget = levelTarget.WithControlPlane(true)
}
```

The smell is in the underlying provider, not in #735.

## Latent impact beyond `target.go`

`resolve` and other read-only commands also call `manager.CurrentTarget()`. `gardenctl resolve shoot --garden X` would hit the same wipe and could return `ErrNoShootTargeted` or wrong scope. Latent bug surface across every consumer.

## Acceptance criteria

- [ ] `manager.Write(t)` followed by `manager.CurrentTarget()` returns `t` (or at least preserves all fields of `t`), regardless of CLI flags.
- [ ] "Move up" semantics for `--garden` still work for the relevant entry-point read (covered by existing tests).
- [ ] Workaround block in `pkg/cmd/target/target.go` (`levelTarget := currentTarget` + switch, introduced in #735) is removed.
- [ ] `resolve` and other read-only commands tested with `--garden` against an unrelated disk target.

petersutter:
**How to categorize this PR?**
/area usability
/kind enhancement
/kind cleanup

Restructures the `target` package so all target mutations flow through a single overlay/merge/validate path, and exposes the resulting target directly from the manager API.

- Direct setters (`TargetGarden/Project/Seed/Shoot/ControlPlane`) and pattern targeting now converge on one `applyOverlay` path that performs merge, validation, enrichment, and access-restriction prompts. The separate `target_builder` is removed.
- `TargetGarden/Project/Seed/Shoot/ControlPlane/MatchPattern` and `UnsetTargetGarden/Project/Seed/Shoot/ControlPlane` return the persisted target. Callers no longer need a follow-up `CurrentTarget()` call, and the merge-wipe workaround in the target command goes away.
- Command-lifetime idempotency moves from the provider into the manager: `CurrentTarget` caches the effective target and all writes update that cache through one `updateTarget` path, so repeated reads no longer reapply CLI flags onto freshly written state.
- The dynamic target provider always merges CLI flags through `merge()`; the fast path that returned `TargetFlags.ToTarget()` for self-contained flags is removed, keeping selector-path clearing and control-plane semantics in one place.
- `--control-plane` becomes tri-state (omitted, true, false). `--control-plane=false` alone clears the targeted control plane via `gardenctl target`. Combining `--control-plane` with path flags but without `--shoot` is rejected.

Fixes #744

**Release note**:
```feature user
`gardenctl target --control-plane=false` now clears the targeted control plane.
```

## Summary by CodeRabbit

* **New Features**
  * Improved target selection and unsetting for gardens, projects, seeds, shoots, and control planes.
  * Added clearer handling of explicit control-plane flag values, including disabling and unsetting.
  * Target commands now provide more accurate success and access-level information.

* **Bug Fixes**
  * Prevented invalid control-plane targets without a shoot or garden.
  * Improved target inheritance, pattern matching, conflict detection, and stale state clearing.
  * Strengthened access-restriction confirmation behavior.

* **Documentation**
  * Clarified selector-path inheritance and the effect of target flags on commands.


pkg/target/manager.go
```diff
@@ -540,76 +575,368 @@ func (m *managerImpl) UnsetTargetControlPlane(ctx context.Context) error {
...
+// call needed for project+seed scope (completeTargetFromShoot fills and
+// canonicalizes values from the shoot object); otherwise project and/or seed
+// are checked individually.
+func (m *managerImpl) validateAndEnrich(ctx context.Context, t Target) (Target, error) {
+	impl, ok := t.(*targetImpl)
+	if !ok {
+		return nil, errors.New("target must be using targetImpl as its underlying type")
+	}
+
+	if impl.Garden == "" {
```
grolu, line 730: assumes Target.Validate() rejects child fields without a garden, but it does not. Removing TargetBuilder also removed its explicit ErrNoGardenTargeted checks

```
➜  darwin-arm64 git:(enh/targeting-ux) ./gardenctl_v2_darwin_arm64 target view
{}
➜  darwin-arm64 git:(enh/targeting-ux) ./gardenctl_v2_darwin_arm64 target project lukas
Error: failed to create garden cluster client: garden name or alias cannot be empty
➜  darwin-arm64 git:(enh/targeting-ux) gardenctl target view
project: lukas
➜  darwin-arm64 git:(enh/targeting-ux)
```

TargetSeed and TargetShoot have the same problem. With kubeconfig linking enabled, the target is written first and only then may the command fail while generating the kubeconfig

pkg/target/target_provider.go
```diff
@@ -147,19 +161,99 @@ func (p *dynamicTargetProvider) Read() (Target, error) {
...
+		gardenName:   combineStringField(cliFlags.GardenName(), patternFlags.GardenName()),
+		projectName:  combineStringField(cliFlags.ProjectName(), patternFlags.ProjectName()),
+		seedName:     combineStringField(cliFlags.SeedName(), patternFlags.SeedName()),
+		shootName:    combineStringField(cliFlags.ShootName(), patternFlags.ShootName()),
+		controlPlane: NewBoolFlag(false),
+	}
+
+	var conflicts []string
+
+	addStringConflict := func(flagName, cliValue, patternValue string) {
```
grolu, line 184: specifying `--garden=<name>` (name alias) no longer works when using together with pattern matching. because garden is compared with identity.
A suitable fix would normalize tf.GardenName() through config.Garden() before comparing it with the pattern-derived canonical identity

pkg/target/manager.go
petersutter: fixed with https://github.com/gardener/gardenctl-v2/pull/766/changes/fd086aeecca6a2eec6d997d012d87a485a58ce80

pkg/target/target_provider.go
petersutter: fixed with https://github.com/gardener/gardenctl-v2/pull/766/changes/b7f57309b60dbff750dd242e102e12d60bb841f4

